PR #908, opened 2014-07-18
Change scaling mode option

HueponiK:
This patch will expand functionality of "letterbox" button.
  Now it is a Combobox. You can choose "Letterbox", or "Stretch", or new option - "Expand". It will expand base resolution so it will fill all the screen, effectively removing black bars from sides or top/bottom, preserving object sizes and aspect ratio.
    That, for example, could be useful  for android port. It would remove side black bars when using original base resolution (SCALE_SCREEN_DIV_3 is still too small for phone and phablets).
    This could also replace SCALE_SCREEN_DIV_X options. We could instead add more options like SCALE_25X, SCALE_3X, etc (maybe even transform them to slider), and leave SCALE_ORIGINAL for esthetes with 4K displays who have very good eyes :P

Normal pixels:
square

1:1.2 pixels:
nonsquare

Please let me know what do you think, should I go for "internal resolution" slider?


HueponiK:
I've decided to kick the can down the road with my small fork - and AUR package, if anyone interested
https://aur.archlinux.org/pkgbase/openxcom-hueponik

PR #927, opened 2014-08-03
make late missions possible

FalkoOXC:
this change allows modded missions to start later in game
the idea ist a mission can start at the earliest in the month where they have a raceweight entry >0
the regions/missions are still selected by random choice
if a mission is selected that is too early i random-select a new one, 
  if that mission is too early i random-select a new one , 
     if that mission is too early i random-select a new one , 
       if that mission is too early i random-select a new one  ... 
i give it 1000 tries (first hit breaks loop so should not hinder vanilla/normal games - vanilla missions have all raceweight: 0: somerace - so these missions always fit)
after 1000 tries  i start to delete the selected missions because its likely that some modder put a STR_NEWMISSION: raceweight: 200: STR_SECTOIDS:100 in a mod so before month 200 it never can take this mission and i would otherwise end in an infinite loop
if the posible mission list is empty it will be newly generated from ruleset
and then i should find new missions and break the loop

### DIFF
--- a/src/Geoscape/GeoscapeState.cpp
+++ b/src/Geoscape/GeoscapeState.cpp
@@ -2163,19 +2163,28 @@ void GeoscapeState::determineAlienMissions(bool atGameStart)
 		// One randomly selected mission.
 		//
 		AlienStrategy &strategy = _game->getSavedGame()->getAlienStrategy();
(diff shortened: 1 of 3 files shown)


Warboy1982:
this won't matter soon enough

PR #959, opened 2014-10-19
Now articles in UFOPaedia can be switched by pressing keys

winterheart:
Now articles in UFOPaedia can be switched by pressing keyPrev and keyNext buttons (left and right arrows by default)


winterheart:
Implemented in 0c758916c30631e5dc1e1df0c6393c54d0fe11e2

PR #963, opened 2014-11-10
Maintenance time is now displayed in the Intercept Window in Geoscape

bslaveboy:


MeridianOXC:
The refuel time on Intercept window is different than on the CraftInfo window for me. Otherwise nice addition.

Issue #152, opened 2012-10-24
Title in 'Select Squad' screen is too long in some lang.


radius75:
![](http://imageshack.us/a/img821/9987/screen003r.png)

Proposal:
- disable _txtTitle->setBig();
- center title
- Y positioning for title (little correct Y for smal text )
- set diff. color for title

cf6b63470e308d3767c440fe0cee10d11d721979 Commit is not ready to Merge
![](http://imageshack.us/a/img222/686/screen006t.png)

PR #965, opened 2014-11-20
Feature Request #152: Change Soldier to Ship in Soldier Screen

BHSDuncan:
This works as follows:
- Right-clicking on one/more soldiers multi-selects the soldier(s), akin to Ctrl+click.
- Shift-right clicking selects a block of soldiers at once, starting from the most-recently selected soldier.
- After the above, left-clicking cycles through the available craft and assigns the soldiers if there's room.  If any of the soldiers are wounded, a warning appears and disallows the selected soldiers from being assigned.
- Crafts with no room or that are out at the moment are not considered for selection.
- If NO soldiers are currently selected via right-click, then left clicking will bring up a soldier's info screen as per normal.

This pull request also includes a change to the TextList class and allows for multi-selection by use of a new boolean flag.  Shift+right click block selection is built in, and right-clicking is akin to ctrl+clicking.

Looking forward to feedback/corrections/bugs/etc.


SupSuper:
This is a really complex interface and TextList is already trying to do too much in one class. No thanks.

PR #971, opened 2014-12-17
Fix gcc-4.9.2 warnings

WorMzy:
Rather than just reporting these individually, I thought I'd have a go at fixing them.


Warboy1982:
whoops, i fixed most of those before i saw the PR, my bad. i'll get that size_t in saveconverter, too.

thanks for the PR and sorry for not noticing it sooner :)

PR #988, opened 2015-02-27
Custom fonts

winterheart:
This pull request introduces option to choose custom fonts without replacing original ones.

New fonts should be placed into data/Languages/ folder. Engine searches this folder for *.dat YAML-file and reads new attribute "name" from it. If YAML don't have this attribute, name will be taken from filename without extension.

New font requires to reload game so this option only accessible from main screen.

I forced to rearrange some options in Video panel to fit all control right.

Aaand some screens!

screen012


winterheart:
Obsoleted by PR #1016.

PR #989, opened 2015-03-01
Simple custom recolors

Yankes:
Support for TFTD aquanauts recoloring.
One drawback it don't have default recoloring for all armors like current version.
Evey one who want use different colors for nationality in his sprite should add line like for `STR_NONE_UC`.

Additional I added option for aliens and civilians too. All code doing it is in last commit, can be skipped.
Only important change its do is move place where seed is loaded form save. This was done to preventing clobbering current seed by using `RNG` in `BattleUnit` constructor.

BTW for completion, do you want add option for recoloring corpses in battlescape and inventory?

### DIFF
--- a/bin/data/Ruleset/Xcom1Ruleset/armors.rul
+++ b/bin/data/Ruleset/Xcom1Ruleset/armors.rul
@@ -2,6 +2,10 @@ armors:
   - type: STR_NONE_UC
     spriteSheet: XCOM_0.PCK
     spriteInv: MAN_0
+    spriteFaceGroup: 6
+    spriteFaceColor: [96, 96, 96, 96, 160, 160, 163, 163] #M0 F0 M1 F1 M2 F2 M3 F3
+    spriteHairGroup: 9
+    spriteHairColor: [144, 144, 164, 164, 245, 245, 166, 166] #M0 F0 M1 F1 M2 F2 M3 F3
     storeItem: STR_NONE
     corpseBattle:
       - STR_CORPSE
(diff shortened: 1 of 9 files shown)


Warboy1982:
i'm pretty sure you can use calls to std::rand() with modulo max value+1 to randomize without skewing the seed, like i do in tile.cpp to randomize smoke/fire patterns on load.

Yankes:
Right, I only do it because I wanted to stick to our RNG, do you want this reverted (and changed to std::rand()) or can it pass?

Warboy1982:
yeah, best to stick to std::rand() for the battleunit constructor, as it gets called on game load anyway, so a save/load would skew it

Yankes:
Ok I rewritten history, There is no RNG in commit anymore.

Warboy1982:
cheers man!

redv:
Why need recolor each frame? You can recolor one time during loading of resources or on demand. Then for each pair ```std::pair<Uint8, Uint8> face_color``` you will have ready to use surface. In fact this is a cache of surfaces. Such pairs is not so much as it seems.

Yankes:
I don't think that coping every surface 8 times will be faster (you faster hit memory limits that cpu ones). My tests shows that you need lot of more operations per pixel to see any performance problems, another thing is Its a lot of easier to expand this with new options, I plan add recoloring for rank and 3 color for race (or random for aliens). But this will be probably for my extended version.

redv:
Is it possible to do a dynamic recoloring? For example: if i want to change the ```face_color``` (race_color, rank_color) in a code, then whether will change looks of the sprite on the screen?

Yankes:
This code don't allow that, its because result is cached before is redrawn on screen. This will cause delay of color change. But I once create more robust version that  allowed recoloring in real time, example:
https://www.youtube.com/watch?v=_G34PHMiBv0
One important part I needed to do was drop of unit sprite cache and draw directly to surface.
Only problem I have with that was I made it too robust, adding custom scripts for recoloring only was overkill. But still I didn't see any big performance drops when I did that. To made significant framerate drop I did 10 times more pixel recolors than was possible in normal OXC play.

PR #992, opened 2015-03-13
Fixed crash caused by serializing doubles imprecisely

cizra:
http://openxcom.org/bugs/openxcom/issues/892 is the crash
https://code.google.com/p/yaml-cpp/issues/detail?id=197 -- this is an issue in yaml-cpp

My code works around the imprecision in yaml-cpp by pre-serializing doubles to hexfloat format. Old savegames remain compatible.


SupSuper:
Nice idea but std::hexfloat is strictly a C++11 feature which we don't support.

cizra:
I changed the code to not require hexfloat (now it works in exactly the same way as in yaml-cpp, only fixed).

PR #993, opened 2015-03-20
Feature: unhardcoded first turn when AI can use grenades.

redv:
Topic: http://openxcom.org/forum/index.php/topic,3219.30.html

Unhardcoded first turn when AI can use grenades or Blaster launcher.

### DIFF
--- a/bin/data/Ruleset/Xcom1Ruleset/vars.rul
+++ b/bin/data/Ruleset/Xcom1Ruleset/vars.rul
@@ -14,3 +14,4 @@ costScientist: 30000
 timePersonnel: 72
 initialFunding: 6000
 alienFuel: STR_ELERIUM_115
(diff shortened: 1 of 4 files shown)


Yankes:
And blast bombs... I think better would be if this was per weapon.

redv:
I found another check for grenades in the getNodeOfBestEfficacy() but i can't find where a check of getTurn() for the launch of Blaster bomb.

Yankes:
Ups.. I thought about `explosiveEfficacy` that have this test too. I mix both (when I last visit this code Warboy didn't add node version).

redv:
Added option for first turn for Blaster launcher.
Thank you)

redv:
Alternative version: https://github.com/SupSuper/OpenXcom/pull/996

PR #994, opened 2015-03-21
Feature: When unit has fatal wounds, his health bar will be flashing.

redv:
Sometimes good soldiers dies because of fatal wounds.
Sometimes good soldiers dies because of fatal wounds because player not know that this soldier has fatal wounds. Because good soldiers cannot speak.

This feature improves the game experience.
When unit has fatal wounds, his health bar will be flashing.
screen020
screen021


IvanDogovich:
I like this! Great playability improvement! Nice Work!

rcreasey:
:+1:

PR #996, opened 2015-03-22
Feature: Unhardcoded first turn when AI can use grenades, launchers etc (alternative version)

redv:
The same as https://github.com/SupSuper/OpenXcom/pull/993 but more general.
Topic: http://openxcom.org/forum/index.php/topic,3219.30.html
This version adds rule for each item: first turn when AI can use this item.


Yankes:
I like this more :D I will probably steal this to my branch if Warboy don't add this :)

Warboy1982:
i preferred the 993 version, can't seem to reopen it at my end

PR #997, opened 2015-04-03
add monthly net funds change info to mfg screen

myk002:
forum discussion:
  http://openxcom.org/forum/index.php/topic,3517.0.html

adds a "NET FUNDS PER MONTH" line to the ManufactureInfoState screen
(where you choose how many items to produce, how many engineers to
allocate to the job, and whether to autosell the product).  This value
will make it clearer how a player's funds will change due to this job.

After discussion in the forums, it was decided that the most useful value
to display was the net change in funds, ignoring other factors such as the
value of used items, the salaries of allocated engineers, or the cost of
used facility space.  The equation used is simply:
  (saleValue - manufactureCost) * itemsPerMonth
where saleValue is the funds you get from selling the item, or 0 if you
are not selling.  manufactureCost is the manufacture/item/cost from the
ruleset, and itemsPerMonth is the number of items the allocated number
of engineers can produce in a month, or the specified number of items to
produce, whichever is less.

Big thank you to forum users Ridan, redv, jgatkinsn, Jstank, and others who contributed valuable feedback that brought this PR to its current form.

Screenshots:
ridan_sell
ridan_no_sell
ridan_0_eng
ridan_1_eng
ridan_2_eng
ridan_4_eng
ridan_50_eng


IvanDogovich:
I think this would be a great addition to the game!  I'm greatly looking forward to it becoming available in an upcoming Nightly! ;)

PR #1000, opened 2015-04-23
Fix: fix "Space Used" in Alien containment

redv:
fix of http://openxcom.org/bugs/openxcom/issues/897

**Before fix**
1 Alien containment after loading. All OK
screen008

2 One muton sent to interrogation. Option "Storage limits" is **off**.
"Space Used" and "Space Available" shows wrong values.
screen009

3 One muton sent to interrogation. Option "Storage limits" is **on**.
"Space Used" and "Space Available" shows wrong values.
screen010

**After fix**
4 One muton sent to interrogation. Option "Storage limits" is **off**.
"Space Used" and "Space Available" shows right values.
screen011

5 One muton sent to interrogation. Option "Storage limits" is **on**.
"Space Used" and "Space Available" shows right values.
screen012

Researched aliens counted in Base::getUsedContainment() https://github.com/SupSuper/OpenXcom/blob/master/src/Savegame/Base.cpp#L1126

PR#1000 :-)

### DIFF
--- a/src/Basescape/ManageAlienContainmentState.cpp
+++ b/src/Basescape/ManageAlienContainmentState.cpp
@@ -50,13 +50,15 @@ namespace OpenXcom
  * @param base Pointer to the base to get info from.
  * @param origin Game section that originated this state.
  */
-ManageAlienContainmentState::ManageAlienContainmentState(Base *base, OptionsOrigin origin) : _base(base), _origin(origin), _sel(0), _aliensSold(0)
+ManageAlienContainmentState::ManageAlienContainmentState(Base *base, OptionsOrigin origin) : _base(base), _origin(origin), _sel(0), _aliensSold(0), _spaceUsed(0), _spaceAvailable(0)
 {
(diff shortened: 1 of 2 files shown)


Warboy1982:
i'm not 1000% sure, but i feel like this was an (evidently botched) attempt to fix another bug

redv:
Researched aliens counted twice.
One time in the constructor ManageAlienContainmentState()
https://github.com/SupSuper/OpenXcom/blob/master/src/Basescape/ManageAlienContainmentState.cpp#L57
second time in Base::getUsedContainment(). Which calls in the ManageAlienContainmentState many times. (hmm. Why it needed?)
https://github.com/SupSuper/OpenXcom/blob/master/src/Savegame/Base.cpp#L1126

I chose to keep a code in the Base::getUsedContainment(). It's OOP-way.
In addition was removed as many code as possible.

Checked exactly 1000 times.
Therefore all should work fine.

Warboy1982:
used this as a template for a slightly different fix

PR #1020, opened 2015-05-17
Read soldier name files from ruleset

myk002:
instead of just globbing the SoldierName dir.  this allows mods to control soldier name pools


Warboy1982:
could we do this at a folder level as well/instead? (ie: filenames with trailing / will be interpreted as "read this folder")

myk002:
@Warboy1982 done.

Warboy1982:
lovely

PR #1021, opened 2015-05-18
Mission triggers

myk002:
allow missions to trigger cutscenes
apply new attributes to cydonia missions for wingame and losegame

noRetreat turned out to be redundant.  it only got checked if (abort || inExitArea == 0), i.e. when you lose the mission.  and once there, the loseCutscene will play regardless, and the CutsceneState will run the endgame logic and drop to the main menu afterwards.


Warboy1982:
mind rebasing this one? the pr seems incredibly messy for whatever reason

myk002:
@Warboy1982  merging master after the slideshow merge seems to have cleaned it up

PR #1029, opened 2015-06-01
Append mission zones instead of overwriting

GixG:
For modding purposes - append new mission zones instead of overwriting - so no need to copy/paste zones/cities from one mod to another...
Modders can use if needed the delete region option for overwrite effect.


FalkoOXC:
so two mods adds missionzones how do you keep the trajectory references and negative textures entries in sync?

GixG:
In case of tampering mission zones (other than cities), use delete
region, re-add region in ruleset.
IMO almost same effort in that case(still one copy/paste needed +
modification)...
But if you want to sum all cities from several mods...

PR #1035, opened 2015-07-30
Only load difficulty from IGLOB.DAT if present

RJPercival:
A bug in the DOS versions of UFO causes the difficulty to be missing
from IGLOB.DAT. Prior to this change, whatever happened to be in memory
past the end of the data vector was loaded as the difficulty. This could
result in aliens having extremely high or negative stats and surprising
end-of-month results.


RJPercival:
Fixed in master.

PR #1039, opened 2015-08-05
Make each game Master (xcom1, xcom2) have own autosave slots.

winterheart:
After switching to TFTD I noticed that openxcom overwrited my old UFO quicksave, autobattlesave and autogeosave :(.
This commit split autosaves for each master game. After that savedir will be little messy, but old autosaves will be still intact and available for loading.

### DIFF
--- a/src/Menu/LoadGameState.cpp
+++ b/src/Menu/LoadGameState.cpp
@@ -55,16 +55,17 @@ LoadGameState::LoadGameState(OptionsOrigin origin, const std::string &filename, 
  */
 LoadGameState::LoadGameState(OptionsOrigin origin, SaveType type, SDL_Color *palette) : _firstRun(0), _origin(origin)
 {
+	std::string curMaster = Options::getActiveMaster();
 	switch (type)
 	{
 	case SAVE_QUICK:
-		_filename = SavedGame::QUICKSAVE;
+		_filename = curMaster + SavedGame::QUICKSAVE;
 		break;
 	case SAVE_AUTO_GEOSCAPE:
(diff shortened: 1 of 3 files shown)


Volutar:
Saves (and screenshots) should be in their own respective game folders. Not just autosaves.

winterheart:
Screens are fine, they have sequence in name, so there no overlapping.
Saves are already filtered by game master, but if user submit same name as savegame from other master, old file will be overwritten, that true. Need to think about it.

killermosi:
Screenshots are global, IMHO.

SupSuper:
Normal saves can't be accidentally overwritten: https://github.com/SupSuper/OpenXcom/blob/master/src/Menu/ListSaveState.cpp#L167

The only reason for this would be organizational purposes. A program doesn't need folders to figure out what's what.

Volutar:
That's why "mess" is the second name of "user" folder?

Volutar:
What if one would need to move tftd saves from one pc to another? how to sort them out? same for screenshots... it needs of some game-hierarchy. It's like every MS file just in "C:\Windows" without subfolders.
Who even invented these folders? Spectrum have had single catalogue with 128 files and it was fine.

winterheart:
Well, after some considerations I decided that all OK in my point of view. Saves with same name are not overwritted, screenshots have own sequencing. Still there only one potential problem - import from original game. But TFTD import not yet implemented, so there no big deal yet.

SupSuper:
Obsoleted by 9e6113f7d8a4d07c310db8c891a6d98672466602

PR #1043, opened 2015-08-13
[bug] Gauss Cannon Ammo is Craft Weapon Ammo.

rcreasey:


SupSuper:
Nope: http://puu.sh/jAi2t/e26ce26a63.png
It's both craft and tank ammo.

PR #1047, opened 2015-08-24
Feature/travis

rcreasey:
Don't merge this until we talk on IRC.  I'll need @SupSuper to make a (free) travis account.

Here's an example of what this does: 
https://travis-ci.org/rcreasey/OpenXcom/jobs/76909855

Note, that this won't replace the nightly build system; this will only run CI on every commit made without actually producing builds for download.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,19 @@
+
+language: cpp
+
+before_install:
+    - if [ $TRAVIS_OS_NAME == osx ]; then brew update && brew install yaml-cpp sdl sdl_gfx sdl_image sdl_mixer --with-flac --with-libmikmod --with-libvorbis --with-static-lib; fi
+    - if [ $TRAVIS_OS_NAME == linux ]; then sudo add-apt-repository ppa:knapsu/openxcom -y; fi
+    - if [ $TRAVIS_OS_NAME == linux ]; then sudo apt-get update && sudo apt-get install -y libsdl1.2-dev libsdl-mixer1.2-dev libsdl-image1.2-dev libsdl-gfx1.2-dev libyaml-cpp-dev libboost-dev; fi 
+      
+script: |
+  cd src && \
+  make -f Makefile.simple
+
+compiler:
+  - clang
+  - gcc
+
+os:
+  - linux
+  - osx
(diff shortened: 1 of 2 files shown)


rcreasey:
I got travis to enable 'osx' builds for this as well.  I'm going try and get that working as well. 

It looks like @SupSuper got Travis setup:
https://travis-ci.org/SupSuper/OpenXcom

So, this pr is good to be accepted so long as the Travis CI service gets enabled on the main github repository.

rcreasey:
Ok, I borrowed some of @JonnyH's travis magic and got cmake working cross platform:
https://travis-ci.org/rcreasey/OpenXcom/builds/77096192

Also, notice this pull request has the awesome 'All checks have passed'.

Here comes the README badge: [Build Status](https://travis-ci.org/rcreasey/OpenXcom)

SupSuper:
What's this about: "This job ran on our legacy infrastructure. Please read our docs on how to upgrade"

LavenderMoon:
If you don't use sudo at all, you can use the new Docker infrastructure for faster compiles and build caching. However, the new infrastructure doesn't support building for Mac.

rcreasey:
Yeah, I'll have a follow up pull request for travis containers:
https://github.com/rcreasey/OpenXcom/tree/feature/travis-container

I'm waiting on them adding a couple more packages to be accepted to their whitelist:
https://github.com/travis-ci/apt-package-whitelist/issues/1000
https://github.com/travis-ci/apt-package-whitelist/issues/1001

And then container builds will be good to go:
https://travis-ci.org/rcreasey/OpenXcom/builds/77102116

As for osx builds on containers, they're only doing those as VMs right now (since you can't containerize a mac just yet).  As you can see from that above linked build, those will just fallback onto the osx infrastructure, while the linux builds are handled by docker infrastructure.

This pull request is good to go though. :+1:

JonnyH:
I actually looked into this a bit earlier and there's already an outstanding apt-whitelist request for libsdl-mixer1.2-dev https://github.com/travis-ci/apt-package-whitelist/issues/715 - apparently they grep for 'suid' so it matches documents and everything....

rcreasey:
Good catch @JonnyH.  I guess I fail at searching.

Let's move the container discussion to this pull request since this one has been merged: https://github.com/SupSuper/OpenXcom/pull/1048

PR #1052, opened 2015-09-18
Fixed type mismatch (signed/unsigned) error in comparsion

NHOrus:
by changing all ints to implementation-defined size_t's


NHOrus:
Note: it's happening in Mod-related files. Should not cause problems with current system and potentially prevent problems with high sprite/sound IDs.

NHOrus:
Breaks stuff horribly: negative IDs are used as keys for custom textures and extra sprites.

PR #1056, opened 2015-09-26
Kneel button toggle

DreamThorn:
Modified the kneel button to indicate whether or not the current unit is kneeling.


DreamThorn:
I have updated this branch to match the latest version.

MeridianOXC:
+1
can we please have this in vanilla? :)

PR #1067, opened 2015-12-04
Dear CMake, you are trying to hard

R1dO:
It is really not necessary to recompile the whole source tree just because the SHA1 hash changes.

So lets compromise.
I give you a sourcefile in which you could endlessly write down SHA1 commits. In return you will only recompile files that did actually change.

Deal?


winterheart:
Here are better way via configure_file directive in CMakefile. I'll figure out what it can be done.

R1dO:
Any help and feedback is always appreciated.

R1dO:
Build works with ``make`` as described in wiki.
Note: Without git-commit SHA substitution, which is consistent with behavior before this PR.
Still need to check building with autotools.

SupSuper:
Not sure why changing a constant would trigger a rebuild, when version.h is only included in a few files, and how would a version.cpp change that.
In any case, I assume this will break all the Git Builders until they're updated accordingly?

R1dO:
Actually it is the definition part `add_definitions( -DOPENXCOM_VERSION_GIT="${git_commit}" )` in cmakelist that triggers the rebuild. Every time it encounters a different SHA1it will update `COMPILE_DEFINITION`, which affects the pre-compiler. 

Unfortunately just setting `set ( OPENXCOM_VERSION_GIT ${git_commit} )` is not enough for the compiler to pickup the variable and set the variable in `version.h` accordingly, hence i let `CMAKE` write it to a separate file.

Although at the moment i am not really happy with it, due to poor choices it probably breaks custom scripts. This PR was needed to see if the travis builds succeed (which they do at the moment), since i can only test the three linux build options (`cmake`, `make`, `autotools`) and not the osx (and windows parts)

Just to prevent misunderstandings. Which git builders did you refer to?

R1dO:
NOTE: only works for `cmake .. && make -j2` kind of builds.
gives error on `make -j2` kind of builds

SupSuper:
#1069

PR #1068, opened 2015-12-06
Fix bug #1139

GixG:
standard data not loaded


SupSuper:
Worked around by loading common data upfront.
If we load everything in Options::init, then any errors will be thrown before the Loading screen is even up, leading to an uglier user experience.

PR #1071, opened 2015-12-10
Describe worktree state in the version string.

R1dO:
Adds "-dirty" when build was based on sources with uncommitted
alterations (hence unknown to bug-hunters).


R1dO:
Build failed, due to travis not being able to pull in yaml-cpp.
But please wait with merging. I might want to add another (small) change.

PR #1072, opened 2015-12-10
Updated package specification files for nightly builds

knapsu:
Updated Ubuntu DEB package specification files to support TFTD changes.


install/debian/control
```diff
@@ -1,16 +1,17 @@
...
 Build-Depends:
 	debhelper (>=9),
+	cmake,
 	libsdl1.2-dev,
 	libsdl-mixer1.2-dev,
-	libsdl-gfx1.2-dev (>= 2.0.22),
+	libsdl-gfx1.2-dev,
 	libsdl-image1.2-dev,
-	libyaml-cpp-dev (>= 0.5.1),
+	libyaml-cpp-dev (>= 0.5),
```
R1dO: This might lead to problems on newer debian/ubuntu systems. Since they use 0.5.2 by default. 
Is there a way you can exclude that version?

knapsu: I am aware of the broken 0.5.2 release in Debian repository. Unfortunately they removed older version from repo so making an exclude "libyaml-cpp-dev (>= 0.5),libyaml-cpp-dev (< 0.5.2)" will prevent openxcom package from building in stable (jessie) and unstable (sid) releases. This files are used on Launchpad PPA where theres a fixed version for yaml 0.5.2+20151114 which is installed over 0.5.2 with openxcom.Doing the 0.5.2 exclude will prevent from using this fixed version and will cause Launchpad builds to fail also. I think it is better to leave this as is because the exclude can hurt more than it tries to solve.
Here I've just synchronized the requirement of "al leasst yaml 0.5" that is described in package specification files for other distributions.

PR #1075, opened 2015-12-28
Align price and quantity to right for better look.

winterheart:
Align to right 3 last columns in Purchase and Sell screens:
screen003
screen004


MeridianOXC:
To be honest, I like the original look better.

JDarthenay:
I agree with Meridian. Is there a forum topic to discuss this?

winterheart:
@JDarthenay, no, this was been my experiment for aligning and better look. Unfortunately, there some problems with widget controls (as I remember, Increase/Decrease buttons cannot be aligned), so this is only demonstration and proof-of-concept.

PR #1077, opened 2016-01-02
Fixed lag issue when research project is completed

wleroux:
Whenever the available research list is produced, the game halts. This
occurs when the new research list is opened as well as when a research
has been completed.

The issue comes from the fact that we check if the research project is
the same name as a unit. Since most research projects are not, an error
occurs as the unit could not be found. The generation of the error
causes the game to lag.

I've changed the code to check for live aliens only when needed as well
as checking if the unit exists instead of getting the unit and ensuring
an error was produced.

### DIFF
--- a/src/Mod/Mod.h
+++ b/src/Mod/Mod.h
@@ -157,6 +157,9 @@ private:
 	/// Loads a ruleset element.
 	template <typename T>
 	T *loadRule(const YAML::Node &node, std::map<std::string, T*> *map, std::vector<std::string> *index = 0, const std::string &key = "type") const;
+	/// Checks if a ruleset element exists.
+	template <typename T>
+	bool hasRule(const std::string &id, const std::map<std::string, T*> &map) const;
 	/// Gets a ruleset element.
 	template <typename T>
 	T *getRule(const std::string &id, const std::string &name, const std::map<std::string, T*> &map) const;
@@ -295,6 +298,8 @@ public:
 	const std::vector<std::string> &getSoldiersList() const;
 	/// Gets commendation rules.
 	std::map<std::string, RuleCommendations *> getCommendation() const;
+	/// Checks if a unit exists.
+	bool hasUnit(const std::string &name) const;
 	/// Gets generated unit rules.
 	Unit *getUnit(const std::string &name) const;
 	/// Gets alien race rules.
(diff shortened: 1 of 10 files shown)


src/Mod/Mod.h
```diff
@@ -157,6 +157,9 @@ class Mod
 	/// Loads a ruleset element.
 	template <typename T>
 	T *loadRule(const YAML::Node &node, std::map<std::string, T*> *map, std::vector<std::string> *index = 0, const std::string &key = "type") const;
+	/// Checks if a ruleset element exists.
+	template <typename T>
```
wleroux: Total Disclaimer: I've never programmed in C++, so I'm not familiar how generics work. We never use the value of the map, so we don't really care about the value type. I thought I'd just stay consistent with other methods.

src/Savegame/SavedGame.cpp
```diff
@@ -1079,18 +1077,11 @@ void SavedGame::getAvailableResearchProjects (std::vector<RuleResearch *> & proj
 					}
 				}
 			}
-			if (!liveAlien && cull)
-			{
-				continue;
-			}
-			else
+
+			if (mod->hasUnit(research->getName()))
```
wleroux: I improperly refactored here, but it turned out to be okay!

The opposite of `!liveAlien && cull` is `liveAlien || !cull)`, not `liveAlien`. However... The block only to sets `cull = false` (which is it already the case if `!cull`.)

SupSuper:
Thanks for the help, but I think I'm just gonna remove the log for now, as there are way too many spots in the code that expect NULL values, so it just ends up being slow and misleading.

wleroux:
Sure thing; this change isn't mutually exclusive to removing the log though... I personally find that having the check be hasUnit(name) more readable than: `if (getUnit(name) != 0)`, `if (!getUnit(name))`, or `if (getUnit(name))`.

In any case, thanks for your work on this project! Great documentation for the contributions! (it didn't work with Visual Studio 2015 though, I had to specifically install Visual Studio 2013.)

SupSuper:
Thanks for the support. I agree that it's more readable, but to fully fix the issue we'd also need hasXXX for every other thing and review all the checks.

As for VS2015, this might help: http://openxcom.org/forum/index.php/topic,3937.msg56803.html#msg56803

wleroux:
Ah, that makes sense. Want me to submit a pull request with no more error logging for getting rules that do not exist?

PR #1078, opened 2016-01-03
Fixed lag issue related to mod rule not existing

wleroux:
Whenever the available research projects are calculated (e.g.: when a research is completed, when the new project button is clicked), we check all of the research projects' names as units. Since the majority of the research projects do not have corresponding units, an error log is generated for every research project.

Having no rule for the name is expected behavior and no error should be generated. This will incidentally fix the lag generated during the available research project calculations (and other places that test existence of a rule this way.)


SupSuper:
Obsolete

wleroux:
Thanks :)

PR #1082, opened 2016-01-16
Fixes buffer overflow

NHOrus:
Deals with things described in http://openxcom.org/bugs/openxcom/issues/1162
by checking for -1 in rankInt before using it as an index.

### DIFF
--- a/src/Mod/Armor.cpp
+++ b/src/Mod/Armor.cpp
@@ -420,7 +420,7 @@ int Armor::getRankColor(int i) const
 {
 	if ((size_t)i < _rankColor.size())
 	{
-		return _rankColor[i];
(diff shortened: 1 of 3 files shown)


Warboy1982:
i'd prefer to default _rankInt to 0 for civs/aliens

NHOrus:
This solution doesn't allow arbitrary modifications in saves to do said buffer overflow.

NHOrus:
Changing -1 to 0 as default rank may break some mods:
https://github.com/SupSuper/OpenXcom/blob/8227eaaf2e16d8ac1da22b35f94caa086990caf3/src/Mod/Armor.cpp#L419-L429

PR #1084, opened 2016-01-23
Option for sharing node values in ruleset

Yankes:
Option for modders to share lot of values between nodes in yaml.
This is work around of lack support of "<<" in yaml-cpp (http://yaml.org/type/merge.html).

This will allow that some nodes could have common values that are same for each node.

``` yaml
items:
  - &refA
    type: TEST_A
    transferTime: 100
    clipSize: 5
  - nodeRef: *refA
    type: TEST_B
  - nodeRef: *refA
    type: TEST_C
  - nodeRef: *refA
    type: TEST_D
```

Now all items have same `transferTime` and `clipSize`.

### DIFF
--- a/src/Mod/AlienDeployment.cpp
+++ b/src/Mod/AlienDeployment.cpp
@@ -131,6 +131,10 @@ AlienDeployment::~AlienDeployment()
  */
 void AlienDeployment::load(const YAML::Node &node)
 {
+	if (const YAML::Node &parent = node["refNode"])
+	{
+		load(parent);
+	}
 	_type = node["type"].as<std::string>(_type);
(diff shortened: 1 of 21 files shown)


SupSuper:
I'd rather this be handled by YAML, since it would be much more flexible. Apparently they have a PR? https://github.com/jbeder/yaml-cpp/pull/300

Yankes:
It would be more flexible, but when its will be available? Another thing that will require switching to C++11 with new yaml-cpp.
I think I could change node name to "<<" to be compatible with yaml definition. When yaml-cpp implements this my code could be removed as not needed anymore.
What you think about this?

PR #1092, opened 2016-03-28
Ufopaedia buildings

GixG:
More information on ufopaedia buildings (storage, personnel cap, radar ranges and chances, etc...)


SupSuper:
Superceded by OXCE.

SupSuper:
@stavrossk OpenXcom Extended is an unofficial branch which adds a lot more power features and mod options that might not be suited for the main repository. You can read more about it here: https://openxcom.org/forum/index.php/board,22.0.html

If a feature is already in OXCE then we'll probably look at their implementation first for consideration into main OXC.

stavrossk:
@SupSuper Thanks, it seems that OpenXcom has  a lot more going for it than I originally thought.

PR #1095, opened 2016-04-03
Fix BR 1202

GixG:
CTD on Ufopedia armor state - it adds empty rows in TextLists, which PR  #1093 doesn't handle adequately.

### DIFF
--- a/src/Interface/TextList.cpp
+++ b/src/Interface/TextList.cpp
@@ -320,12 +320,6 @@ void TextList::addRow(int cols, ...)
 		}
 		if (cols > 0)
 			txt->setText(va_arg(args, wchar_t*));
-		else
-		{
-			wchar_t d[1];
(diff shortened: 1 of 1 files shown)


src/Interface/TextList.cpp
```diff
@@ -307,7 +318,14 @@ void TextList::addRow(int cols, ...)
...
 			txt->setSmall();
 		}
-		txt->setText(va_arg(args, wchar_t*));
+		if (cols > 0)
+			txt->setText(va_arg(args, wchar_t*));
+		else
+		{
+			wchar_t d[1];
+			mbstowcs(d, "", 1);
+			txt->setText(d);
```
SupSuper: I assume you're doing the same as `txt->setText(L"");`
But the whole `else` is redundant anyways since a `Text` is always created with an empty string.

PR #1102, opened 2016-05-01
Move borderless windows

JDarthenay:
Gives the possibility to move borderless windows by providing its coordinates.
I had to resort to a workaround to force SDL moving window when width and height of game screen are unchanged.

Players such as me having two screens need that feature to move window to the secondary screen.


JDarthenay:
What is the problem with this pull request? Any chance to see an error message or something?

JDarthenay:
It should be fine now. If there is a problem with my code, please let me know.

SupSuper:
Nice idea. Some users would also like to be able to set "windowedModePositionX/Y" from the Options menu. Maybe combine both options into one, and make it a generic "Window Position" option? It would also make the code more concise and readable.

src/Engine/Screen.cpp
```diff
@@ -342,6 +362,20 @@ void Screen::resetDisplay(bool resetVideo)
 		}
 #endif
 		Log(LOG_INFO) << "Attempting to set display to " << width << "x" << height << "x" << _bpp << "...";
+
+		// Workaround for window not moving when keeping same size
+		if (_screen != 0 && _screen->w == width && _screen->h == height && Options::borderless)
+		{
+			if (width != 320)
```
SupSuper: Consider using constants here, eg. Screen::BASE_WIDTH and Screen::BASE_HEIGHT

src/Interface/TextEdit.cpp
```diff
@@ -497,8 +506,14 @@ void TextEdit::keyboardPress(Action *action, State *state)
 			break;
 		default:
 			Uint16 key = action->getDetails()->key.keysym.unicode;
-			if (((_numerical && key >= L'0' && key <= L'9') ||
-				(!_numerical && ((key >= L' ' && key <= L'~') || key >= 160))) &&
+			// If constraint is "signed numeric", need to check:
```
SupSuper: The if is getting too big, it should be moved into its own function (eg. isValidChar(key)) and the conditions logically broken into multiple if's.

src/Interface/TextEdit.h
```diff
@@ -75,6 +76,8 @@ class TextEdit : public InteractiveSurface
 	void setVerticalAlign(TextVAlign valign);
 	/// Sets the text edit to numerical input.
 	void setNumerical(bool numerical);
+	/// Sets the text edit to signed numerical input.
+	void setSignedNumerical(bool numerical);
 	/// Sets the text edit's color.
```
SupSuper: setNumerical and setSignedNumerical could just be combined into a setConstraint(constraint)

src/Menu/OptionsVideoState.cpp
```diff
@@ -528,6 +542,22 @@ void OptionsVideoState::btnLockMouseClick(Action *)
 }
 
 /**
+ * Ask user where he wants to root screen.
+ * @param action Pointer to an action.
+ */
+void OptionsVideoState::btnFixedBorderlessPosClick(Action *)
+{
+	if ((SDL_GrabMode)_btnFixedBorderlessPos->getPressed())
```
SupSuper: What's this cast doing here?

JDarthenay:
@SupSuper Would it really be user friendly to force him/her to adopt same coordinates for the two modes?

src/Interface/TextEdit.h
JDarthenay: That was what I wanted to do but I was afraid you would consider this a too big impact on existing code. Well, I should not have been afraid.

src/Menu/OptionsVideoState.cpp
JDarthenay: It seems to me I copied this piece of code from somewhere else (and then changed only button variable name), so probably I duplicated a useless cast.

src/Menu/SetBorderlessRootState.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+	add(_window, "confirmVideo", "mainMenu");
+	add(_btnOk, "confirmVideo", "mainMenu");
+	add(_btnCancel, "confirmVideo", "mainMenu");
+	add(_txtTitle, "confirmVideo", "mainMenu");
+	add(_txtWindowPositionNewX, "confirmVideo", "mainMenu");
+	add(_txtWindowPositionNewY, "confirmVideo", "mainMenu");
+	add(_edtWindowPositionNewX, "confirmVideo", "mainMenu");
+	add(_edtWindowPositionNewY, "confirmVideo", "mainMenu");
+
```
JDarthenay: I am really not sure the arguments I have put here to those eight `add` statements are correct.

JDarthenay:
@SupSuper Do I also have to update the `pch.h` file as Yankes suggest if [here](http://openxcom.org/forum/index.php/topic,4452.msg64167.html#msg64167)?

JDarthenay:
Not good yet, this is triggering confirm windows when not changing display options, I need to fix this.

JDarthenay:
I have done a few testing, it should be fine now. Limitations are:
- not able to move a resizable window after startup without changing resolution, the workaround I had to use for other modes leads to infinite resizes loop with a resizable window.
- If user moves Window out of sight, he has to wait 15 seconds before the changes are cancelled and he get his window back.
- some confirmation changes may seem unnecessary, but we need them as we cannot know with SDL 1.2 if new coordinates user is giving are different from current window coordinates.
- if user sets his window on a secondary screen and the next time he plays OpenXcom he doesn't have this screen (or he change screens config or whatever) the window will be out of screen, user will have to kill game with `taskkill /F` and change options.cfg by himself. This could be counterbalanced by adding a center window key shorcut, buy I was not able to decide which key shortcut would be accurate.

JDarthenay:
@SupSuper Hello again, what do I have to do more so you can accept this pull request?

SupSuper:
Sorry I'll test it out this weekend at most. We're busy people and don't want too many disruptions before the 2.0 release, so pull requests can take a while to get merged.

src/Engine/Screen.cpp
```diff
@@ -342,6 +337,22 @@ void Screen::resetDisplay(bool resetVideo)
 		}
 #endif
 		Log(LOG_INFO) << "Attempting to set display to " << width << "x" << height << "x" << _bpp << "...";
+
+		// Workaround for window not moving when keeping same size
```
JDarthenay: @SupSuper Why did you remove this work around? As a result window is not moved immediately unless it is resized...

SupSuper: It causes crashes on some systems: https://github.com/SupSuper/OpenXcom/commit/9d44544c204b597e59bbd23221b7c48f589623c2

PR #1103, opened 2016-05-04
CMAKE MacOS app bundle

neuralshock:
Hi,

This pull request has the goal of trying to build an OpenXcom app bundle for MacOS using only brew-installed software and the XCode Command Line Tools (thus avoiding the installation of a 7GB bundle from the App Store)

The build doesn't work yet, the binary cannot find the bin_data_dirs inside Resources.

**TODO**
- [x] Rename app bundle to OpenXcom.app
- [x] Set app bundle icon
- [x] bin_data_dirs (common, standard, TFTD, UFO) should be copied to the "Resources" folder of the app bundle
- [ ] Working build


SupSuper:
Good luck!

src/CMakeLists.txt
```diff
@@ -759,8 +760,14 @@ if ( FORCE_INSTALL_DATA_TO_BIN )
...
-install ( TARGETS openxcom ${install_dest} DESTINATION ${CMAKE_INSTALL_BINDIR} )
+if ( APPLE )
+	add_executable ( OpenXcom ${application_type} ${openxcom_src} )
+	install ( TARGETS OpenXcom ${install_dest} DESTINATION ${CMAKE_INSTALL_BINDIR} )
+else ()
+	add_executable ( openxcom ${application_type} ${openxcom_src} )
+	install ( TARGETS openxcom ${install_dest} DESTINATION ${CMAKE_INSTALL_BINDIR} )
+endif ()
+
 # Extra link flags for Windows. They need to be set before the SDL/YAML link flags, otherwise you will get strange link errors ('Undefined reference to WinMain@16')
```
winterheart: Eh, why did you rename openxcom executable? Double work here and below in file.

neuralshock: AFAIK the executable must have the same name as the bundle, otherwise it will not work.
I'm trying to keep it consistent with the existing Mac build via Xcode, as well (which uses that naming convention).

Should I do it differently?

winterheart: Well, maybe better to change on Xcode part...

Can you please try generate xcode project from CMake (`cmake -G Xcode`), maybe it would working solution?

neuralshock:
It fails, since it requires the full XCode suite (which this pull request is trying to avoid).

neuralshock:
An update, some time later: the CMake-generated Xcode project doesn't work properly.
I don't have the time to work on this anymore so I'll close the pull request. 

Thanks for the support, anyway!

PR #1104, opened 2016-05-07
CTRL+G not properly taken into account when used from the Options Video State.

JDarthenay:
Moving a line downward because CTRL+G is not properly taken into account when used from the Options Video State.
`_states.back()->handle(&action);` has to be called after the "ctrl-g" key is processed.
Note: not sure how is handled `Options::debug`.


JDarthenay:
@SupSuper Hello, is there a problem with this pull request? Maybe the reason why I did it is not clear? I hope you will apply this bug fix. Do I have to put a patch on the forum so other players can test it?

PR #1105, opened 2016-05-12
Cmake improvements

winterheart:
* Changed *.cpp files, affected by last commit, fixes building.
* Stripped header files from CMakeLists.txt, they don't need to be here

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,668 +1,344 @@
 set ( root_src
-  dirent.h
-  fmath.h
   lodepng.cpp
-  lodepng.h
   main.cpp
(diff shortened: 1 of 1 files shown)


SupSuper:
Out of curiosity, why aren't the header files required? I've always seen them specified in CMakeLists.txt files.

winterheart:
Well, CMake is smart enough to include headers to build process autmagically. Even if header has been changed, cmake will rebuild all files that include them. That works by design.

Only one concern to add headers to CMakeLists.txt is improper import to IDEs. That was true for old QtCreator, but recent IDEs can import cmake-project properly.

PR #1110, opened 2016-06-01
Use KDialog when in a KDE desktop session

atlimit8:


R1dO:
This PR merely changes the bias from a GTK dialog towards a specific implementation of a QT dialog.

If the error dialog must be tuned more towards the users current desktop i believe it is better to query the "XDG_CURRENT_DESKTOP" variable and act depending on that outcome.
You can use [this](https://specifications.freedesktop.org/menu-spec/latest/apb.html) site for values that are most likely encountered.

atlimit8:
Actually, the first condition checks for `KDE_SESSION_UID` environment variable which is only set during a KDE session. If the user is not running KDE, Zenity and the rest (including KDialog) are checked like normal.

Order in active KDE desktop environment:
 * KDialog - __conditional__
 * Zenity
 * KDialog - __unconditional__
 * GDialog
 * Xdialog

Order otherwise:
 * Zenity
 * KDialog
 * GDialog
 * Xdialog

To be clear, if the user has both KDialog and Zenity but is not running this from a *KDE session*, **_Zenity_** with be chosen instead of KDialog because `getenv("KDE_SESSION_UID") && system("which kdialog 2>&1 > /dev/null") == 0` will be false.

To be fair it is easy to think that this _unconditionally_ changes the bias if you miss or don't know what `getenv("KDE_SESSION_UID")` checks.

---
I did research `XDG_CURRENT_DESKTOP` environment variable and it is not necessarily set, especially older versions and uncommon DEs. See http://unix.stackexchange.com/questions/116539/how-to-detect-the-desktop-environment-in-a-bash-script for example.

---
That said, I am not apposed to adding detection other DEs with the understanding that detection is not guaranteed and the choice of dialog utilities may not be obvious. Note that KDE, Gnome, and Unity are the three major desktop environments.

R1dO:
I might have been a bit strong on words when mentioning the bias but my main issue is that this check is very centred towards a single desktop environment.

I do see your problem when users of a QT desktop (like KDE) have installed some software which brings zenity in as a dependency. I am just more a fan of solutions that benefits more than just the big 3 (which can even be different than the ones you've pointed out depending on the source of usage statistics).

I am aware of the downsides of the variable I've mentioned, it just feels more natural to query one variable (which usage has probably risen since the article from 2014 when it was still relative young) instead of checking multiple environmental variables based on which DE is targeted.

That being said, I do appreciate your efforts for user experience consistency even for use cases which we all try to avoid triggering when coding ;)

PR #1111, opened 2016-06-05
Don't reset _configFolder if it isn't necessary

Manuel-K:
On Linux the -config/-cfg argument is ignored if the -user argument isn't used at the same time.

This fixes the problem.


R1dO:
Your selling yourself a bit short here,

You've also managed to fix this behaviour for all systems that have a notion of separate configuration directories (which is everything except apple and windows according to CrossPlatform).

Thanks for the fix though, should simplify my debug scripts.

PR #1113, opened 2016-07-02
Check for upper tile's existence

sfalexrog:
In Pathfinding.cpp, the aboveDestination pointer is not checked before being used. Judging by the surrounding code, this is probably an oversight more than anything else (belowDestination pointer is actually checked).


SupSuper:
Obsolete

PR #1116, opened 2016-11-01
Fix invalid _targetVoxel when LOF is blocked

franciscobacona:
Given the following situation:

	 012345678
	0   ____
	1  |    |
	2  |    |B
	3  |____|
	4      A

A takes an autoshot at B, which is visible, but LOF is blocked by a
wall to the north.

The first shot removes the northern wall at (6, 4):

	012345678
	1  ____
	2 |    |
	3 |    |B
	4 |___ |
	5     A

However, the second and third shots fly NW, perpendicular to the path to
B, removing the walls at (2, 2) and (3, 1)!

	012345678
	1   ___
	2      |
	3 |    |B
	4 |___ |
	5     A

In fact, the projectiles are flying towards (-16, -16, -24) due to
commit 62ec237f08530295a419e67a5a6b4336abc0df04.

This patch fixes this behavior by aiming at the middle of the target
tile in this case.

### DIFF
--- a/src/Battlescape/ProjectileFlyBState.cpp
+++ b/src/Battlescape/ProjectileFlyBState.cpp
@@ -229,7 +229,8 @@ void ProjectileFlyBState::init()
 			{
 				if (!_parent->getTileEngine()->canTargetUnit(&originVoxel, targetTile, &_targetVoxel, _unit))
 				{
-					_targetVoxel = Position(-16,-16,-24); // out of bounds, even after voxel to tile calculation.
+					// target nothing, targets the middle of the tile
+					_targetVoxel = Position(_action.target.x*16 + 8, _action.target.y*16 + 8, _action.target.z*24 + 12);
 				}
 			}
 		}


franciscobacona:
Here is a vanilla savegame displaying the behavior. Take an autoshot with the selected soldier at the soldier nearest to him and observe the plasma shots head mysteriously towards grid origin.

[_targetVoxel test.sav.txt](https://github.com/SupSuper/OpenXcom/files/563549/_targetVoxel.test.sav.txt)

hellrazor4223:
This will not solve the problem. These cases should be handled as No Line Of Fire incidents.
The issue is in this case that the enemy can see you but can not target you properly, such a case was already report here: http://openxcom.org/bugs/openxcom/issues/1297

The solution is actually quite simple:
diff -ruN ../src_orig/Battlescape/Projectile.cpp ../src_new/Battlescape/Projectile.cpp
--- ../src_orig/Battlescape/Projectile.cpp	Sun Oct 30 04:09:41 2016
+++ ../src_new/Battlescape/Projectile.cpp	Thu Oct 27 20:14:25 2016
@@ -121,6 +121,7 @@
 {
 	Tile *targetTile = _save->getTile(_action.target);
 	BattleUnit *bu = _action.actor;
+	if (_targetVoxel.z < 0) return V_EMPTY;
 	
 	int test;
 	if (excludeUnit)

From volutar, already tested this and it works :)

franciscobacona:
@hellrazor4223 

Ah, I should have checked the bug tracker first. Thanks for posting the patch. Happy to see a fix is coming.

PR #1120, opened 2016-11-09
Fix opengl program attach

JonnyH:
Without this, *all* opengl shaders would fallback to 'no shader'....

I'll be available for a public flogging Friday.

### DIFF
--- a/src/Engine/OpenGL.cpp
+++ b/src/Engine/OpenGL.cpp
@@ -338,7 +338,6 @@ static GLuint createShader(GLenum type, const char *source)
 	GLint compileSuccess;
 	glGetShaderiv(shader, GL_COMPILE_STATUS, &compileSuccess);
 	glErrorCheck();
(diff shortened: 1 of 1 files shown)


src/Engine/OpenGL.cpp
```diff
@@ -362,11 +361,21 @@ static GLuint createShader(GLenum type, const char *source)
...
+	{
+		glAttachShader(glprogram, fragmentshader);
+		glErrorCheck();
+	}
 }
 
 void OpenGL::set_vertex_shader(const char *source)
 {
 	vertexshader = createShader(GL_VERTEX_SHADER, source);
+	if (fragmentshader)
```
dopplershift: shouldn't that be `if (vertexshader)`?

JonnyH: Indeed. The flogging shall continue.

PR #1126, opened 2016-11-28
Use C++ limits, instead of rolling our own.

uqs:
While here, remove redundant casts on return types, remove parentheses
around return, as it's not a function call. This makes the overall style
more consistent.

Taken from: FreeBSD ports
Original author: madpilot@FreeBSD.org


SupSuper:
C++ 64-bit limits are only available on C++11.

PR #1127, opened 2016-11-28
Remove a FreeBSD specific ifdef and a redundant _WIN32 ifdef

uqs:
It's fine to also check in /usr/share/openxcom under FreeBSD, no need to
special case this. Update the docs to reflect that this is being checked
always (this was an omission from the Linux docs previously.)

Trim some trailing tabs while here.


uqs:
Travis timed out, how can I force a re-run?

README.md
```diff
@@ -120,7 +120,7 @@ Data folders:
 - /Users/Shared/OpenXcom
 - . (the current directory)
 
-### Linux
+### Linux / \*BSD
```
SupSuper: We normally just use "Unix-like" to avoid offending any sensibilities.

uqs: Reverted that line

PR #1131, opened 2017-01-18
fixing misleading spacing resolving compile error with gcc

slowriot:
without this fix, this fails to compile with gcc due to -Werror=misleading-indentation:

```
src/Engine/OpenGL.cpp: In member function 'void OpenXcom::OpenGL::refresh(bool, unsigned int, unsigned int, unsigned int, unsigned int, int, int, int, int)':
src/Engine/OpenGL.cpp:154:3: error: this 'while' clause does not guard... [-Werror=misleading-indentation]
   while (glGetError() != GL_NO_ERROR); // clear possible error from who knows where
   ^~~~~
src/Engine/OpenGL.cpp:155:2: note: ...this statement, but the latter is misleadingly indented as if it is guarded by the 'while'
  clear();
  ^~~~~
```

### DIFF
--- a/src/Engine/OpenGL.cpp
+++ b/src/Engine/OpenGL.cpp
@@ -1,4 +1,4 @@
-// This file was copied from the bsnes project. 
+// This file was copied from the bsnes project.
 
 // This is the license info, from ruby.hpp:
 
@@ -21,7 +21,7 @@
 
 namespace OpenXcom
 {
-	
+
 bool OpenGL::checkErrors = true;
 
 std::string strGLError(GLenum glErr)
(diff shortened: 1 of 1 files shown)


Manuel-K:
Where did you get the `-Werror=misleading-indentation` from? It does not seem to be enabled by default…

Anyhow, `src/Engine/OpenGL.cpp` has a mixture of space and tab indentation. It's not pretty, but it is not the only file that mixes both styles. There are 937 files containing tabs and 240 files that contain at least one line that is indented with multiple spaces.

slowriot:
@Manuel-K it may not be the only file that has a mess of tabs and spaces, but it's the only one that prevents compilation.

That error is enabled by default in Debian's 6.3.0 - I received this building the repo "out of the box" with a standard Debian Testing machine.  Therefore this fix, at least, is necessary to build on Debian Testing with GCC 6.

PR #1133, opened 2017-02-22
Fix for http://openxcom.org/bugs/openxcom/issues/641

kristomu:
This fixes the bug/improvement request of http://openxcom.org/bugs/openxcom/issues/641.

It handles the first part of the issue by truncating too long lines, and the second part of the issue by replacing the greedy word wrap algorithm with a minimum raggedness algorithm.

Finally, it adds some minor changes not related to either: fixes a typo in the US text (` instead of ' was rendering as ?) and replaces magic constants like \xA0 for nonbreaking spaces with names like TOK_NBSP.

I'm using English TFTD, so I can't recreate the exact same dialog boxes as in issue 641, but see images linked below:

https://imgur.com/a/fF8r4
https://imgur.com/a/tkCan
https://imgur.com/a/ucnkv

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -91,6 +91,8 @@ if ( NOT OPENGL_FOUND )
     "Continuing building without OpenGL support."
     )
   add_definitions(-D__NO_OPENGL)
+else ()
+    include_directories ( ${OPENGL_INCLUDE_DIR} )
 endif ()
 
 # Read version number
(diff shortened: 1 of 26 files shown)


SupSuper:
I've been going very back and forth on this, but in the end while I appreciate the effort, this adds a whole lot of complexity to the text rendering without really solving any issues (the linked bug is more of a general problem with UI layout and 320x200 rendering than wrapping limitations). Plus the simpler dumber wordwrapping looks more vanilla.

I like your minor fixes and readability improvements though so I'm gonna cherry-pick those.

kristomu:
The idea with solving the second issue is that if you have

ALVO: LOCAL DE TERROR-
6

then a word wrap that seeks to minimize the difference in the length of each line will give

ALVO: LOCAL DE
TERROR-6

or something to that extent, because the second line is no longer very short compared to the first. So I think it would still solve that problem. But I can see the point about it not looking vanilla any longer.

Would having a nonbreaking hyphen character be a good simpler way of handling the problem? Or would that still be too complex?

PR #1134, opened 2017-02-22
Unified (and fixed) chart labeling algorithm

MeridianOXC:
Country lines were drawn incorrectly.
When there was negative score, the *full* range was compared against only 8/9 of available chart space.

Technical:
1. the "check" is calculated against (0, range)
2. but "range" includes the portion below X axis too, so grids-- is not necessary/is wrong
3. the shift below X axis is done later (correctly)


scyt4l3:
no more following of this thread

PR #1139, opened 2017-03-11
Save options.cfg ordered and with empty line at EOF

l0b0:
This makes the file much easier to compare between changes, and also makes it easier to version control.

Based on (OK, completely ripped off) [code by @hersh](https://github.com/jbeder/yaml-cpp/issues/169#issuecomment-219763811).


Warboy1982:
looks good to me, any objections?

PR #1148, opened 2017-05-26
Vs2017 compile

hartmark:
Used these precompiled dependancies:
  https://openxcom.org/forum/index.php/topic,3937.msg80377.html#msg80377

Had some warnings that I have now fixed.

Release compile now also have level 1 warnings (same as debug already had) now we don't have thousands of warnings.

Also we should have a link on this page to the precompiled dependancies:
http://ufopaedia.org/index.php/Compiling_with_Microsoft_Visual_C%2B%2B_(OpenXcom)

### DIFF
--- a/src/Engine/FlcPlayer.cpp
+++ b/src/Engine/FlcPlayer.cpp
@@ -21,7 +21,9 @@
  * Based on http://www.libsdl.org/projects/flxplay/
  */
 #ifdef _MSC_VER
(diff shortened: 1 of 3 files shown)


hartmark:
@SupSuper any comments on this PR?

SupSuper:
@hartmark I think defining YAML_CPP_DLL breaks older versions of yaml-cpp where the macro worked differently. The rest looks fine though.

PR #1150, opened 2017-06-05
Add Reworked Research Functions

hellrazor4223:
Made the code m
ore understandable.
Alive aliens and items with "getOneFree" will now stay researchable for the player as long as there is a undiscovered "getOneFree" or "unlocks", in the ruleset.

The alive alien check is now much more understandable and basically allows unhardocding of STR_LEADER_PLUS and STR_COMMANDER_PLUS, the explizit check for these are not needed anymore.

I tested the code against vanilla and my own mod. Works :)

### DIFF
--- a/src/Battlescape/Pathfinding.cpp
+++ b/src/Battlescape/Pathfinding.cpp
@@ -243,6 +243,17 @@ bool Pathfinding::aStarPath(Position startPosition, Position endPosition, Battle
 	return false;
 }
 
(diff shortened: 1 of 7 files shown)


hellrazor4223:
Takes care of the following Bug report:
https://openxcom.org/bugs/openxcom/issues/1215
https://openxcom.org/bugs/openxcom/issues/1157

PR #1151, opened 2017-06-05
volutars door patch

hellrazor4223:
volutars door patch
just so we have the code on github.
Source: http://volutar.myds.me/05_doors.patch

- patching door walking and proper TU indication (currently doors are displaying wrong TUs for big units) http://openxcom.org/bugs/openxcom/issues/995


hellrazor4223:
Sorry need more branches

PR #1153, opened 2017-06-10
generalized research handling

hellrazor4223:
Please review and consider merging thanks.

### DIFF
--- a/src/Savegame/SavedGame.cpp
+++ b/src/Savegame/SavedGame.cpp
@@ -1139,6 +1139,8 @@ void SavedGame::getAvailableResearchProjects (std::vector<RuleResearch *> & proj
 	std::vector<std::string> researchProjects = mod->getResearchList();
 	const std::vector<ResearchProject *> & baseResearchProjects = base->getResearch();
 	std::vector<const RuleResearch *> unlocked;
+
+	// Create list of pointers for all unlocks for already discovered research topics and handle errors
 	for (std::vector<const RuleResearch *>::const_iterator it = discovered.begin(); it != discovered.end(); ++it)
 	{
 		for (std::vector<std::string>::const_iterator itUnlocked = (*it)->getUnlocked().begin(); itUnlocked != (*it)->getUnlocked().end(); ++itUnlocked)
@@ -1146,84 +1148,71 @@ void SavedGame::getAvailableResearchProjects (std::vector<RuleResearch *> & proj
 			unlocked.push_back(mod->getResearch(*itUnlocked, true));
 		}
 	}
-	for (std::vector<std::string>::const_iterator iter = researchProjects.begin(); iter != researchProjects.end(); ++iter)
+
+	// Create list of research topics available to be researched on the given base, shown in research menu
+	for (std::vector<std::string>::const_iterator itModResearchList = researchProjects.begin(); itModResearchList != researchProjects.end(); ++itModResearchList)
 	{
-		RuleResearch *research = mod->getResearch(*iter);
-		if (!isResearchAvailable(research, unlocked, mod))
+		RuleResearch *research = mod->getResearch(*itModResearchList);
+
+		// Check for dependencies, unlocked list, debug and possible  initial stop conditions
+		if (!isResearchAvailable(research, unlocked))
 		{
 			continue;
 		}
-		std::vector<const RuleResearch *>::const_iterator itDiscovered = std::find(discovered.begin(), discovered.end(), research);
 
-		bool liveAlien = mod->getUnit(research->getName()) != 0;
-
-		if (itDiscovered != discovered.end())
+		// Check if "requires" are satiesfied
+		if (!research->getRequirements().empty())
 		{
-			bool cull = true;
-			if (!research->getGetOneFree().empty())
-			{
-				for (std::vector<std::string>::const_iterator ohBoy = research->getGetOneFree().begin(); ohBoy != research->getGetOneFree().end(); ++ohBoy)
-				{
-					std::vector<const RuleResearch *>::const_iterator more_iteration = std::find(discovered.begin(), discovered.end(), mod->getResearch(*ohBoy));
-					if (more_iteration == discovered.end())
-					{
-						cull = false;
-						break;
-					}
-				}
-			}
-			if (!liveAlien && cull)
+			if(!isResearched(research->getRequirements()))
 			{
 				continue;
 			}
-			else
+		}
+
+		// Check which topics are researched and
+		// how to handle reresearchable topics to keep them on the list
+		if(isResearched(research->getName()))
+		{
+			bool removeFromMenu = true;
+
+			// Hide hidden topics / fake researches
+			// One time use only
+			// see STR_PLASMA_CANON_DEP1 STR_PLASMA_CANON_DEP2 as example
+			if(research->getCost() == 0 || research->getPoints() == 0)
 			{
-				std::vector<std::string>::const_iterator leaderCheck = std::find(research->getUnlocked().begin(), research->getUnlocked().end(), "STR_LEADER_PLUS");
-				std::vector<std::string>::const_iterator cmnderCheck = std::find(research->getUnlocked().begin(), research->getUnlocked().end(), "STR_COMMANDER_PLUS");
+				continue;
+			}
 
-				bool leader ( leaderCheck != research->getUnlocked().end());
-				bool cmnder ( cmnderCheck != research->getUnlocked().end());
-
-				if (leader)
+			if(!research->getGetOneFree().empty())
+			{
+				if(!isResearched(research->getGetOneFree()))
 				{
-					std::vector<const RuleResearch*>::const_iterator found = std::find(discovered.begin(), discovered.end(), mod->getResearch("STR_LEADER_PLUS"));
-					if (found == discovered.end())
-						cull = false;
+					removeFromMenu = false;
 				}
-
-				if (cmnder)
+			}
+			if(!research->getUnlocked().empty())
+			{
+				if(!isResearched(research->getUnlocked()))
 				{
-					std::vector<const RuleResearch*>::const_iterator found = std::find(discovered.begin(), discovered.end(), mod->getResearch("STR_COMMANDER_PLUS"));
-					if (found == discovered.end())
-						cull = false;
+					removeFromMenu = false;
 				}
-
-				if (cull)
-					continue;
+			}
+			if(removeFromMenu)
+			{
+				continue;
 			}
 		}
 
+		// Check if this topic is already being researched on the current base
 		if (std::find_if (baseResearchProjects.begin(), baseResearchProjects.end(), findRuleResearch(research)) != baseResearchProjects.end())
 		{
 			continue;
 		}
+
+		// Check for needed item on current base
 		if (research->needItem() && base->getStorageItems()->getItem(research->getName()) == 0)
 		{
 			continue;
-		}
-		if (!research->getRequirements().empty())
-		{
-			size_t tally(0);
-			for (size_t itreq = 0; itreq != research->getRequirements().size(); ++itreq)
-			{
-				itDiscovered = std::find(discovered.begin(), discovered.end(), mod->getResearch(research->getRequirements().at(itreq)));
-				if (itDiscovered != discovered.end())
-				{
-					tally++;
-				}
-			}
-			if (tally != research->getRequirements().size())
-				continue;
 		}
 		projects.push_back (research);
 	}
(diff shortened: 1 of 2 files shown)


src/Savegame/SavedGame.cpp
```diff
@@ -1270,56 +1237,37 @@ bool SavedGame::isResearchAvailable (RuleResearch * r, const std::vector<const R
...
+        if(!r->getUnlocked().empty())
+        {
+                if(isResearched(r->getUnlocked()))
+                {
+                        allUnlocks = true;
+                }
+        }
+        if(allGetOneFree && allUnlocks)
+        {
+                return false;
```
MeridianOXC: If you completely remove the code on lines 1251-1270... the algorithm will still do the same, no? The code on lines 1164-1186 will take care of this... with a small exception on line 1175, which I don't understand.

src/Savegame/SavedGame.cpp
```diff
@@ -1149,60 +1149,41 @@ void SavedGame::getAvailableResearchProjects (std::vector<RuleResearch *> & proj
...
+                {
+                        bool removeFromMenu = true;
+                        if(!research->getGetOneFree().empty())
+                        {
+                                if(!isResearched(research->getGetOneFree()))
+                                {
+                                        removeFromMenu = false;
+                                }
+                        }
+                        if(research->getCost() != 0 && !research->getUnlocked().empty())
```
MeridianOXC: Why checking the cost here? In some mods (e.g. PirateZ) there are perfectly intentional researchable topics with cost = 0.

src/Savegame/SavedGame.cpp
```diff
@@ -1149,60 +1149,41 @@ void SavedGame::getAvailableResearchProjects (std::vector<RuleResearch *> & proj
...
-						cull = false;
-				}
-
-				if (cull)
-					continue;
-			}
 		}
-
+                std::vector<const RuleResearch *>::const_iterator itDiscovered = std::find(discovered.begin(), discovered.end(), research);
+                if(itDiscovered != discovered.end())
```
MeridianOXC: This is OK, but it can also be written nicer as: if (isResearched(research)) and line 1164 could be removed.

src/Savegame/SavedGame.cpp
```diff
@@ -1149,60 +1149,41 @@ void SavedGame::getAvailableResearchProjects (std::vector<RuleResearch *> & proj
 	for (std::vector<std::string>::const_iterator iter = researchProjects.begin(); iter != researchProjects.end(); ++iter)
 	{
 		RuleResearch *research = mod->getResearch(*iter);
+
```
MeridianOXC: I still think there is code duplication between the getAvailableProjects and isResearchAvailable functions... see the comment on the bottom, at line 1269. Maybe it would help to see it more clearly, if you would remove the isResearchAvailable function completely and move its code directly here... then it will become more obvious where the duplication is.

hellrazor4223: I do not think that removing the isResearchAvailable function is a good idea.
I already tried to set the checks for "dependencies" and "requires" within the same function before, but i will recheck if it is possible with the same functionality or not. Even thou having some initial condition checks, before starting to build the push_back is good, but it might not need two function.

src/Savegame/SavedGame.cpp
hellrazor4223: That seems feasible, i was more focused on the rest :D

src/Savegame/SavedGame.cpp
hellrazor4223: The isResearchAvailable function is called before we start building the push_back list, so having some initial checks before we do that might be good, also debug handling is in it. Even thou yes the code might still act the same if we would remove the said lines, i will test this.

src/Savegame/SavedGame.cpp
hellrazor4223: Checking the cost is a quick way to remove fake research topics like STR_PLASMA_CANON_DEP1 and STR_PLASMA_CANON_DEP2 from the menulist we are building at that point.
Looking at how these fake research topics are defined in the ruleset, we need to make a explizit check for them, they need to be culled from the list. We can extend the check to also include getPoints() !=0 to make we only get those. The fake research topics usually only use "requires", "dependencies", "unlocks". They also usually do not have a name string for language translation set (->extraStrings). 
Points, cost and naming string would be the appropriate checks for them. What do you think?

hellrazor4223: Unsupported languages will have no naming string defined, so getCost and getPoints will have to suffice.

hellrazor4223:
I implemented some changes and also some requests of @MeridianOXC 
also commented the code for more readability.
Thanks for the review @MeridianOXC

src/Savegame/SavedGame.cpp
```diff
@@ -1270,56 +1255,30 @@ bool SavedGame::isResearchAvailable (RuleResearch * r, const std::vector<const R
 	{
 		return false;
 	}
-	std::vector<std::string> deps = r->getDependencies();
-	const std::vector<const RuleResearch *> & discovered(getDiscoveredResearch());
-	bool liveAlien = mod->getUnit(r->getName()) != 0;
```
JonnyH: Now nothing uses the argument "mod" - maybe remove it from the args list? (This causes a warning in gcc, as it's "often" a mistake if someone ended up not using a variable they were expecting to)

src/Savegame/SavedGame.cpp
```diff
@@ -1139,92 +1139,77 @@ void SavedGame::getAvailableResearchProjects (std::vector<RuleResearch *> & proj
...
 		}
 
+                // Check for reresearchable topics and keep them on the list
+                // as long as they have something to discover
+                if(isResearched(research->getName()))
+                {
+                        bool removeFromMenu = true;
+                        // Hide hidden topics / fake researches
+                        // Do not do these checks on them and remove them from list in any case
+                        if(research->getCost() != 0 && research->getPoints() != 0)
```
JonnyH: can this be flattened into:

if (research->getCost() == 0 || research->getPoints() == 0)
{
  continue;
}
if (!research->getOneFree().empty())
.... etc etc ...

src/Savegame/SavedGame.cpp
hellrazor4223: I removed it, it is indeed here no longer needed.

src/Savegame/SavedGame.cpp
hellrazor4223: For more readability. - Done :)

hellrazor4223:
Thanks @JonnyH 
I put in your two changes.

src/Savegame/SavedGame.cpp
```diff
@@ -1139,92 +1139,81 @@ void SavedGame::getAvailableResearchProjects (std::vector<RuleResearch *> & proj
...
+		// Check which topics are researched and
+		// how to handle reresearchable topics to keep them on the list
+		if(isResearched(research->getName()))
+		{
+			bool removeFromMenu = true;
+
+			// Hide hidden topics / fake researches
+			// One time use only
+			// see STR_PLASMA_CANON_DEP1 STR_PLASMA_CANON_DEP2 as example
+			if(research->getCost() == 0 || research->getPoints() == 0)
```
MeridianOXC: This is a better place for checking fake topics as before, but... as I said before, there are valid (not fake) research topics with cost == 0 in mods. Can we at least check cost == 0 && points == 0 (instead of cost == 0 || points == 0)? That should be a lot safer.

src/Savegame/SavedGame.cpp
```diff
@@ -1264,62 +1253,36 @@ void SavedGame::getAvailableProductions (std::vector<RuleManufacture *> & produc
...
+		if(!isResearched(r->getDependencies()))
 		{
-			return true;
+			return false;
 		}
 	}
 
-	for (std::vector<std::string>::const_iterator iter = deps.begin(); iter != deps.end(); ++ iter)
+	// Skip reresearchable topics with "getOneFree" and "unlocks" once they are completly exhausted
+	if(!r->getGetOneFree().empty() && (!r->getUnlocked().empty()))
```
MeridianOXC: This is wrong, I think you didn't understand exactly Johnny's IRC comment. For example, if the topic can be researched multiple times (like a live alien for example), and it has several "getOneFree" but doesn't have any "unlocks", the code would not handle such situation. The previous code you wrote was OK... suggestion from Johnny was also meant well, but wasn't implemented correctly here.

src/Savegame/SavedGame.cpp
MeridianOXC: Also Johnny's suggestion to move it even more up seems nice. If the topic is a fake topic, we don't need to do many other things above. The check could go even completely on the top.

src/Savegame/SavedGame.cpp
```diff
@@ -1139,92 +1139,81 @@ void SavedGame::getAvailableResearchProjects (std::vector<RuleResearch *> & proj
...
 		{
 			continue;
 		}
-		std::vector<const RuleResearch *>::const_iterator itDiscovered = std::find(discovered.begin(), discovered.end(), research);
-
-		bool liveAlien = mod->getUnit(research->getName()) != 0;
 
-		if (itDiscovered != discovered.end())
+		// Check if "requires" are satiesfied
+		if (!research->getRequirements().empty())
```
MeridianOXC: Here (and on similar places)... why do you check the condition if (!research->getRequirements().empty()) ? You can remove this condition completely and if(!isResearched(research->getRequirements())) would still work properly.

src/Savegame/SavedGame.cpp
```diff
@@ -1139,92 +1139,81 @@ void SavedGame::getAvailableResearchProjects (std::vector<RuleResearch *> & proj
...
-					if (found == discovered.end())
-						cull = false;
+					removeFromMenu = false;
 				}
-
-				if (cmnder)
+			}
+			if(!research->getUnlocked().empty())
+			{
+				if(!isResearched(research->getUnlocked()))
```
MeridianOXC: I see the problem now... this is why you checked for fake research. That's however not the correct solution. The if(!isResearched(research->getUnlocked())) is wrong. Instead of checking whether the unlocks are already discovered, you only need to check if they are unlocked. The check if they are discovered is "too strong".

src/Savegame/SavedGame.cpp
MeridianOXC: Also, we need to see fake research too... this is important when upgrading mods. When you upgrade the mod and research tree changes, you need to be able to research these fake topics as well, to fix your research tree mid-game. This is current vanilla functionality and should not be removed.

src/Savegame/SavedGame.cpp
hellrazor4223: The old code also was to clunky and in this case also not working.
I now only checks for research topics which do have both, and removes them if they are fully satiesfied.
I will try to see if i can also make it work with the other conditions.
Unless we just remove it, since the other code already takes care of it.

src/Savegame/SavedGame.cpp
hellrazor4223: The check is only performed against already researched ones, to hide them from the list then. In the case you describe the should be showing up on the list, as before. Needs testing thou, do you have a testcase?

I did check with && before and @JonnyH suggested the change.
But your right checking for a logical and is what we want here.
Even thou this here is giving me a headache:
http://en.cppreference.com/w/cpp/language/operator_logical

src/Savegame/SavedGame.cpp
hellrazor4223: Its more clear what we actually do there in the first place, also why call the string comparison function, if it is not needed.
In our case here it does return "true" if the given topic is empty.
But better to have a explizit check at least in my eyes.

src/Savegame/SavedGame.cpp
hellrazor4223: The discovered list is what is saved in the savegame. All other checks, like "requires" and "dependencies" etc.. are performed against it. No reason to differentiate here.

MeridianOXC: Well, if you leave it like this, you will be able to research for example Sectoid Soldiers forever. Just start a new game, put 10 Sectoid Soldiers into your Alien Containment and then research them one by one, without researching anything else. You can see that you can research all 10 of them, even though only the first one can give you something... other 9 are useless and should not be shown. This is also how current code works, and what is correct in my opinion.

src/Savegame/SavedGame.cpp
MeridianOXC: I don't see any string comparison function... what do you mean? There is only a check for empty vector, same as what you were doing explicitly here.

hellrazor4223: Do not get me wrong, in my eyes it makes very clear what the code does or what not, so i would like to keep them around.

src/Savegame/SavedGame.cpp
MeridianOXC: If the code below on line 1195 will be fixed, this check will not be necessary at all. See changes I sent you via PM.

src/Savegame/SavedGame.cpp
hellrazor4223: Research Alien Origins, then the Soldier will disappear from the list.
Would be nice to actually have it removed directly once Alien Origins is available for research in the list.

src/Savegame/SavedGame.cpp
MeridianOXC: Yes, the old code was wrong here too. I made the changes in my commit already, which I sent you via PM.

hellrazor4223: I was thinking, with the additional stop conditions gone, nothing would really be in the way to get rid of the "isResearchAvailable" function.
I will change and test it. And make a seperate commit instead of forcepushing.

src/Savegame/SavedGame.cpp
MeridianOXC: That's not correct, it must disappear without researching Alien Origins, unlocking origins is enough.

src/Savegame/SavedGame.cpp
MeridianOXC: I have removed isResearchAvailable already in my commit... please have a look at that.

src/Savegame/SavedGame.cpp
hellrazor4223: Maybe checking for the unlock list here might be needed. I will try i out.
Also i only stated what is was doing and yes that their is indeed a issue.

MeridianOXC: Exactly... and I did that change in my commit too, in case you want to save some time.

hellrazor4223: I tested you code in your commit and it is not correctly working.
The topics with multiple unlocks are getting pulled to early from the list, breaking vanilla main story research.

hellrazor4223:
I could possible live with aliens being pulled late, but probably the rest not.
Also @MeridianOXC seemed to have found even more errors in other code segments.
"addFinishedResearch" for example.

hellrazor4223:
All my code has already been absorbed and slightly modified by @MeridianOXC in his PR: https://github.com/SupSuper/OpenXcom/pull/1155

PR #1156, opened 2017-07-10
Added Italian translation of main components of the engine

AresAndy:
The title says it 👓 .. This adds some Italian translation to the engine. I'll update it when other locale files are found


SupSuper:
https://www.transifex.com/openxcom/openxcom/language/it/

PR #1158, opened 2017-07-23
Added damage range min and max for finer control of said range.

ryz:
I've been working on a rebalancing mod for both UFO and TFTD, so a finer control about the lower and upper bounds of the damage range would be very useful to me. Let me know what you think!


src/Battlescape/TileEngine.cpp
```diff
@@ -1066,8 +1066,13 @@ BattleUnit *TileEngine::hit(Position center, int power, ItemDamageType type, Bat
...
 	{
 		int dmgRng = type == DT_HE ? Mod::EXPLOSIVE_DAMAGE_RANGE : Mod::DAMAGE_RANGE;
-		int min = power * (100 - dmgRng) / 100;
-		int max = power * (100 + dmgRng) / 100;
+
+		//int min = power * (100 - dmgRng) / 100;
+		//int max = power * (100 + dmgRng) / 100;
+
+		int min = Mod::DAMAGE_RANGE_MIN;
+		int max = Mod::DAMAGE_RANGE_MAX;
```
MeridianOXC: How is this finer control? Instead of different damage ranges for explosives and other weapons, we now have just one damage range...

MeridianOXC: Wait a moment, this is completely wrong... weapon power is not even included in the calculation. Or am I missing something?

ryz: Oops! You're right! Sorry, my bad. I should not commit while i'm in transit!

Anyway, my point is that I would like to adjust the upper and lower bounds of weapon damage spread indepentently, which is currently not possible.

MeridianOXC: OK, if you need more control, you can override ranges per weapon in OXCE+. You can also make 2-dice rolls instead of 1-dice.

SupSuper:
Superceded by OXCE

PR #1160, opened 2017-07-30
fix behavior of fixed weapons in inventory screen

myk002:
ensure fixed weapons can't be removed/added by ctrl-clicking or via
inventory template operations

This ensures tank-like units that show up on the equip screen (e.g. attack dogs in Final Mod Pack) can't have their fixed weapons replaced before battle.


MeridianOXC:
Nice fixes. I did it almost identically in OXCE+ 2 years ago :)

PR #1169, opened 2018-01-05
Only award revivedSoldier commendation once per medikit state

R1dO:
When an unconscious soldier has multiple wounds, but is revived after the 1st heal action, the current situation increase the revivedSoldier on each subsequent healing action (assuming the player did not exit the medikit).

This PR ensures it will only be awarded once.

> note:
>  This PR deliberately does not change the different medikit behavior regarding simulant and heal 
>  actions. Which can be observed in the attached mod+savegame:
>  When selected soldier operates medikit on unconscious one it is standing over.
>  * Using simulants to revive the soldier will exit immediately if target regains consciousness
>  * Using heal will not exit immediately if soldier regains consciousness

[commendationsHealMedikitExample.zip](https://github.com/SupSuper/OpenXcom/files/1605231/commendationsHealMedikitExample.zip)

### DIFF
--- a/src/Battlescape/MedikitState.cpp
+++ b/src/Battlescape/MedikitState.cpp
@@ -119,7 +119,7 @@ MedikitButton::MedikitButton(int y) : InteractiveSurface(30, 20, 190, y)
  * @param targetUnit The wounded unit.
  * @param action The healing action.
  */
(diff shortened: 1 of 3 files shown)


R1dO:
Issue has been fixed by @Warboy1982 in commit 2eaaf484c8580564497962f159e17918cd2307a8
Hence closing the PR

PR #1170, opened 2018-01-20
Append to the tooltip string the productions and researches.

GixG:
If mouse hovering over a laboratory, append to the tooltip string the base's current researches. If mouse hovering over a wokshop, append to the tooltip string the base's current manufacturing.


SupSuper:
This is way too much information for a tooltip.

PR #1172, opened 2018-02-04
Reduced compilation time and simplified include hierarchy.

qrr:
**In short**

Reduced compilation time by eliminating unnecessary includes via forward declaring the types most interfaces required for serialization, and reducing inclusion scopes.

This was specifically tested with CMake & GCC & LLVM/Clang under Debian Stretch; no ill effects were observed either during build, or during gameplay.

The changes result in approximately 15% improvement on fast machines.

**In depth**

Compilation time was extremely slow on an older P3 laptop, and was in the multiple-minute range with 4 cores using a modern CPU: the former was 70+ minutes (!), and the latter was over 4 minutes.

According to GCC profiling, most of the time is taken up by parsing and template instantiation, which indicates nothing unusual is going on.

Examining the include hierarchy, I've come to the conclusion that multiple quite expensive headers are pulled into separate compilation units; these are:
        - YAML-CPP's all-inclusive header.
        - SDL's all-inclusive header.

The serialization interfaces are replicated consistently across the codebase, thus I've altered their includes first. Unfortunately, since YAML-CPP requires the template specialization in the following files to be present at the site of usage, the definition thereof had to be included in the files which serialized them in some manner.

Affected files:
- Unit.h
- Position.h
- RuleRegion.h
- Unit.h
- Texture.h

These files were given a new ".hpp" file, which contains their serialization related template specialization. This also indirectly affects multiple files, which depended upon the definition of these specializations. These pull in the ".hpp" directly.

Furthermore, two files received .cpp files to house their own definitions or I/O operators:
- MissionStatistics.h's definitions were externalized.
- Position.h's I/O operators were externalized. The definitions were left in place to facilitate inlining. (Though optimization wasn't the goal in this session.)

I've also taken the liberty of altering the include order in the touched files: the STL includes follow and internal and third party includes, to avoid polluting them with their contents. This revealed missing includes, which were added. The include order in the files touched is:
 - Internal project includes.
 - The specific .hpp includes mentioned previously.
 - External library includes.

These do not appear to be against the code conventions posted [here](https://openxcom.org/forum/index.php/topic,109.0.html).

Other notes:
 - Operated on the general assumption that any of the affected files are not singly used, or won't stay so, ie. at least two files will use the raw header files. Otherwise, it would not make sense to separate the serialization specialization.
 - Replaced SDL.h with SDL_stdinc.h, and SDL_video.h, etc. where appropriate: no need to pull in the entire SDL library just for type declarations.

Central includes:
  - State.h is extremely central, understandably due to the architectural decisions.
 - Mod.h is the one of the most central includes, and it pulled in Unit.h, which itself contained the serializing specialization.
 - Game.h is also high on the list.
 - TextButton.h is central too. and it pulls in InteractiveSurface, which itself pulls in Surface.h and State.h.
 - Options.h is central, which makes OptionInfo central as well.

I'll let the numbers speak for themselves, as measured on my development machine:
 - This version compiles in approximately 3 minutes 32 seconds.
 - The original compiles in approximately 4 minutes 9 seconds.
 - Thus, this is approximately a 15% improvement. I've yet to measure on my older laptop, but if the numbers hold, this is already a good 10 minutes faster there.
 -  The above numbers are from singular measurements, but the dispersion wasn't large enough to warrant serious statistics.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,6 +4,8 @@ cmake_minimum_required ( VERSION 2.8.5 )
 set ( CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake/modules" )
 include(GNUInstallDirs)
 
+set (CMAKE_CXX_STANDARD 11)
+
 option ( DEV_BUILD "Development Build. Disable this for release builds" ON )
 option ( BUILD_PACKAGE "Prepares build for creation of a package with CPack" ON )
 option ( ENABLE_WARNING "Always show warnings (even for release builds)" OFF )
--- a/src/Basescape/SoldierDiaryMissionState.cpp
+++ b/src/Basescape/SoldierDiaryMissionState.cpp
@@ -31,6 +31,8 @@
 #include "../Savegame/MissionStatistics.h"
 #include "../Savegame/BattleUnitStatistics.h"
 
+#include <yaml-cpp/yaml.h>
+
 namespace OpenXcom
 {
 
@@ -130,7 +132,7 @@ void SoldierDiaryMissionState::init()
 		missionId = 0;
 	}
 	int daysWounded = missionStatistics->at(missionId)->injuryList[_soldier->getId()];
-	
+
 	_lstKills->clearList();
 	_txtTitle->setText(tr(missionStatistics->at(missionId)->type));
 	if (missionStatistics->at(missionId)->isUfoMission())
@@ -145,7 +147,7 @@ void SoldierDiaryMissionState::init()
 	_txtDaylight->setText(tr("STR_DAYLIGHT_TYPE").arg(tr(missionStatistics->at(missionId)->getDaylightString())));
 	_txtDaysWounded->setText(tr("STR_DAYS_WOUNDED").arg(daysWounded));
 	_txtDaysWounded->setVisible(daysWounded != 0);
-	
+
 	int kills = 0;
 	bool stunOrKill = false;
 
--- a/src/Battlescape/AIModule.cpp
+++ b/src/Battlescape/AIModule.cpp
@@ -16,8 +16,6 @@
  * You should have received a copy of the GNU General Public License
  * along with OpenXcom.  If not, see <http://www.gnu.org/licenses/>.
  */
-#include <climits>
-#include <algorithm>
 #include "AIModule.h"
 #include "../Savegame/BattleItem.h"
 #include "../Savegame/Node.h"
@@ -35,6 +33,11 @@
 #include "../Mod/Mod.h"
 #include "../Mod/RuleItem.h"
 #include "../fmath.h"
+
+#include <yaml-cpp/yaml.h>
+
+#include <climits>
+#include <algorithm>
 
 namespace OpenXcom
 {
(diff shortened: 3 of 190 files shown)


MeridianOXC:
It will be a lot of fun merging this into OXCE and OXCE+ :(

qrr:
I may be able to help with that - up until you've mentioned it, I wasn't aware of the [project](https://github.com/Yankes/OpenXcom/tree/OpenXcomExtended). I appear to have arrived to this fantastic project in interesting times, as the mod site is less than functional - no need to elaborate, I've read up on the situation - and thus haven't really discovered anything in the 'scene' besides FMP, and that only through ... now that I check, apparently _your_ Let's Play videos. :)

I would really not like to make things more difficult for any developers working on related projects I may not know about - the core goal here is to make hacking speedier. The underlying reason is that I'd like to be able to work on the performance 'issues' **on** the devices which _have_ those issues - which are devices I don't believe particularly many people would care about, such as my vintage laptop.

I really welcome any suggestions or ideas how to make this more 'conformant', and if you can point me towards relevant design / architectural differences between this codebase - which I assume to be the origin, based on the apparent Subversion migration - and the common targets of the community, please do so.

Aside, I'm unable to reproduce the issues tripping up the Travis build on iOS, nor can I see the reason for them besides a potentially broken YAML-CPP dependency upstream.

sfalexrog:
For what it's worth, the build issues on macOS stem from the fact that recent yaml-cpp requires c++11 to build. You can probably work around that by checking for yaml-cpp version in CMakeLists.txt file and adding "-std=c++11" to the OpenXcom target conditionally.

qrr:
Indeed, 0.6.0 and the recent 0.6.1 requires C++11. Quite rational - I initially didn't want to touch the CMake configuration as that might mess things for others way more than necessary, but here it is. I don't believe making it conditional on YAML-CPP version would make much sense, since - as Travis demonstrates - it can make the codebase more difficult to maintain. It is really easier to board the already venerable C++11 ship.

This opened a can of worms, however, so it is on to play whack-an-error with Travis - though I may look into proper conditions to reproduce the issue on my machine.

Here's to making an allegedly scary pull request even more so!

MeridianOXC:
FYI, I believe SupSuper is deliberately keeping C++11 out of this project.

sfalexrog:
Seems like a tough thing to do when one of the dependencies requires C++11 to even *use* it. How bad would it be if OpenXcom had its own version of yaml-cpp that worked but required Boost instead?
It seems that OpenXcom does something similar for its linux builds (it checks out yaml-cpp 0.5.1 and builds from source). Maybe it could do something similar for OS X? (Though having yaml-cpp stuck at one particular version would be really sad; I remember seeing non-trivial performance improvements with later versions, and there might have been a memory leak that got fixed later)

But then again, it would only solve problems for travis builds, and fixing travis builds just for travis builds' sake seems counterproductive. End users compiling from source with recent enough yaml-cpp will run into the same problems.

But having C++11 enabled for CMake project would require at the very least adapting some code, and would surely lead to some C++11 creeping in. Right now I can't see an easy way out.

SupSuper:
Postponed til after v2.0

PR #1173, opened 2018-02-15
Graphs screen hotkeys

RangelReale:
F1: enable all that have activity >= 1
F2: enable all that have activity >= 7


RangelReale:
After testing for I while, I say this didn't help very much, so I am closing it.

PR #1176, opened 2018-03-12
WIP: Another highly cocky attempt to mass-manipulate codebase

NHOrus:
Included are fruit of clang-tidy and manual refactoring of four enums
More enum classes to follow, much magic was slain.


NHOrus:
OpenXCom does not support and is not written in C++11

PR #1177, opened 2018-03-13
Annual cleanup and warning fix

NHOrus:
Updated LodePNG version, did some things to enums and readability in AIModule

### DIFF
--- a/src/Battlescape/Pathfinding.cpp
+++ b/src/Battlescape/Pathfinding.cpp
@@ -668,13 +668,14 @@ bool Pathfinding::isBlocked(Tile *tile, const int part, BattleUnit *missileTarge
 		}
 	}
 	// missiles can't pathfind through closed doors.
-	if (missileTarget != 0 && tile->getMapData(part) &&
-		(tile->getMapData(part)->isDoor() ||
-		(tile->getMapData(part)->isUFODoor() &&
-		!tile->isUfoDoorOpen(part))))
+	{ TilePart tp = (TilePart)part;
+	if (missileTarget != 0 && tile->getMapData(tp) &&
+		(tile->getMapData(tp)->isDoor() ||
+		(tile->getMapData(tp)->isUFODoor() &&
+		!tile->isUfoDoorOpen(tp))))
 	{
 		return true;
-	}
+	}}
 	if (tile->getTUCost(part, _movementType) == 255) return true; // blocking part
 	return false;
 }
--- a/src/Battlescape/Pathfinding.cpp
+++ b/src/Battlescape/Pathfinding.cpp
@@ -623,9 +623,9 @@ bool Pathfinding::isBlocked(Tile *tile, const int part, BattleUnit *missileTarge
 	}
 	if (part == O_FLOOR)
 	{
-		BattleUnit *unit = tile->getUnit();
-		if (unit != 0)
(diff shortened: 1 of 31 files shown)


src/Battlescape/TileEngine.cpp
```diff
@@ -1588,7 +1588,7 @@ int TileEngine::verticalBlockage(Tile *startTile, Tile *endTile, ItemDamageType
 	int block = 0;
 
 	// safety check
-	if (startTile == 0 || endTile == 0) return 0;
+	if (!startTile || !endTile == 0) return 0;
```
SupSuper: This is wrong

src/Interface/NumberText.cpp
```diff
@@ -183,11 +183,11 @@ NumberText::NumberText(int width, int height, int x, int y) : Surface(width, hei
 		_borderedChars[i] = new Surface(5, 7);
 		// give it a border
 		// this is the "darker" shade that goes in the corners.
-		for (int x = 0; x <= 2; x += 2)
```
SupSuper: Would be better to keep the "x" and "y" since they refer to coordinates. Maybe "xx" and "yy"?

SupSuper:
In general:
- Best not to touch the null pointer checks, wouldn't wanna cause an error and accidentally introduce a subtle bug.
- When renaming variables due to "shadowing", make sure the new names are still meaningful.

src/Battlescape/Pathfinding.cpp
```diff
@@ -668,13 +668,14 @@ bool Pathfinding::isBlocked(Tile *tile, const int part, BattleUnit *missileTarge
 		}
 	}
 	// missiles can't pathfind through closed doors.
-	if (missileTarget != 0 && tile->getMapData(part) &&
-		(tile->getMapData(part)->isDoor() ||
-		(tile->getMapData(part)->isUFODoor() &&
-		!tile->isUfoDoorOpen(part))))
+	{ TilePart tp = (TilePart)part;
```
SupSuper: Why the extra bracket?

NHOrus: Scoping. So `tp` will go away after closing bracket.

src/Battlescape/TileEngine.cpp
NHOrus: Fixed by amending commit

src/Menu/VideoState.cpp
```diff
@@ -317,13 +317,13 @@ static introSoundEffect introSoundTrack[] =
...
 {
-	Mod *mod;
+	Mod *_mod;
 	Music *m;
 	Sound *s;
 	int trackPosition;
 	FlcPlayer *_flcPlayer;
 
-	AudioSequence(Mod *mod, FlcPlayer *flcPlayer) : mod(mod), m(0), s(0), trackPosition(0), _flcPlayer(flcPlayer)
+	AudioSequence(Mod *mod, FlcPlayer *flcPlayer) : _mod(mod), m(0), s(0), trackPosition(0), _flcPlayer(flcPlayer)
```
NHOrus: This was bad idea, should have gone with `AudioSequence(Mod *_mod, FlcPlayer *flcPlayer)` instead

NHOrus: Fixed

PR #1179, opened 2018-04-08
fix bug where mercy or nike cross is erroneously awarded for instance…

jgatkinsn:
…s where someone gets all the kills and someone else gets all the stuns in a single mission (supposed to be all aliens killed or stunned by single person not both in same mission)

### DIFF
--- a/src/Battlescape/DebriefingState.cpp
+++ b/src/Battlescape/DebriefingState.cpp
@@ -486,11 +486,11 @@ DebriefingState::DebriefingState() : _region(0), _country(0), _positiveScore(tru
 					soldierAlienStuns++;
 				}
 			}
-			if (aliensKilled && aliensKilled == soldierAlienKills && _missionStatistics->success == true)
+			if (aliensKilled && aliensKilled == soldierAlienKills && _missionStatistics->success == true && aliensStunned == 0)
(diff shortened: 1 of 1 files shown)


SupSuper:
What happens if a soldier kills+stuns every alien on a mission?

jgatkinsn:
Ran test.  No medal as expected.  I can handle this two ways:  implement a new award for taking out all the aliens with whatever means necessary, or I may have to edit the commendations.  I was trying to avoid changing the mod too much, since I didn't write it or the code to go with it.  I'm just trying to fix obvious errors based on their descriptions.  Should I see if can get in contact with Shoes?

MeridianOXC:
I like Ivan's proposal. Nike cross for "eliminate (=stun or kill) all" and Mercy cross for "stun all".

SupSuper:
I approve. Let's go with door number 3.

While you're there, please change "if (aliensKilled)" to "if (aliensKilled == 0)", as mixing ints and bools is kinda confusing.

jgatkinsn:
Okay.  I'll do that. Both door number 3 and fix aliensKilled (aliensStunned).

jgatkinsn:
Done, and done

PR #1180, opened 2018-04-22
adding faction granularity when reviving target units

jgatkinsn:
This fixes issues with awarding angel cross medal in Commendations where it's suppose to only award when reviving fellow soldier (as in one of your own).  The statistic was too generic and all encompassing for all revived units.  This fix will also allow awards to be made for reviving civilians and hostile units.

### DIFF
--- a/src/Battlescape/MedikitState.cpp
+++ b/src/Battlescape/MedikitState.cpp
@@ -234,7 +234,18 @@ void MedikitState::onHealClick(Action *)
 			if (!_revivedTarget)
 			{
 				_targetUnit->setTimeUnits(0);
-				_action->actor->getStatistics()->revivedSoldier++;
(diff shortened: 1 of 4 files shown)


src/Battlescape/MedikitState.cpp
```diff
@@ -234,7 +234,18 @@ void MedikitState::onHealClick(Action *)
 			if (!_revivedTarget)
 			{
 				_targetUnit->setTimeUnits(0);
-				_action->actor->getStatistics()->revivedSoldier++;
+                if(_targetUnit->getOriginalFaction() == FACTION_PLAYER)
```
SupSuper: Use tabs instead of spaces

jgatkinsn:
Will do.

PR #1183, opened 2018-05-01
Normalize line endings

MeridianOXC:


jgatkinsn:
Should .rul files also be added to the explicit conversion list alongside .cpp and .h?

PR #1186, opened 2018-07-01
Split "MIDI" music format to "MIDI (GM.CAT)" and "MIDI (MID files)"

Player701:
This pull request has been created to implement [feature request #1397](https://openxcom.org/bugs/openxcom/issues/1397) filed on the OpenXcom bug tracker.

Currently, if "MIDI" is selected as the music format, songs from the GM.CAT file will be used if it exists, otherwise .MID files will be used. This means that if GM.CAT is present, .MID files cannot be used at all unless GM.CAT is removed from the SOUND subfolder of the game data directory. This also affects mods: if a mod provides music replacement via .MID files, they will not be played back if GM.CAT is present.

This change allows the user to choose between MIDI music from the GM.CAT file (if it exists) and separate .MID files. Besides the fact that the Steam version of UFO Defense comes with both GM.CAT and .MID files, this change will also benefit mod users (see above).

A minor issue with this change is that it appears to be impossible to detect which type of MIDI music is currently playing, since the current format is queried by a call to SDL_mixer. Also, the names of the new options in the combo box could probably be edited to be more user-friendly.

### DIFF
--- a/src/Engine/Options.h
+++ b/src/Engine/Options.h
@@ -33,7 +33,7 @@ enum KeyboardType { KEYBOARD_OFF, KEYBOARD_ON, KEYBOARD_VIRTUAL };
 /// Savegame sorting modes.
 enum SaveSort { SORT_NAME_ASC, SORT_NAME_DESC, SORT_DATE_ASC, SORT_DATE_DESC };
 /// Music format preferences.
-enum MusicFormat { MUSIC_AUTO, MUSIC_FLAC, MUSIC_OGG, MUSIC_MP3, MUSIC_MOD, MUSIC_WAV, MUSIC_ADLIB, MUSIC_MIDI };
+enum MusicFormat { MUSIC_AUTO, MUSIC_FLAC, MUSIC_OGG, MUSIC_MP3, MUSIC_MOD, MUSIC_WAV, MUSIC_ADLIB, MUSIC_GM, MUSIC_MIDI };
 /// Sound format preferences.
 enum SoundFormat { SOUND_AUTO, SOUND_14, SOUND_10 };
 /// Video format preferences.
--- a/src/Menu/OptionsAudioState.cpp
+++ b/src/Menu/OptionsAudioState.cpp
@@ -126,8 +126,8 @@ OptionsAudioState::OptionsAudioState(OptionsOrigin origin) : OptionsBaseState(or
 	musicText.push_back(L"MOD");
 	musicText.push_back(L"WAV");
 	musicText.push_back(L"Adlib");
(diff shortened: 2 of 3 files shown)


stavrossk:
Glad to see you guys are still working on this game.

Player701:
Thanks, but I am not a member of the OpenXcom development team. I do, however, hope too that they are still working on this project and could find some time to review my proposed changes.

stavrossk:
@Player701 It's good to see you are still invested even though you are not a member of the OpenXcom team. X-Com is the best game ever made and I don't wanna see this project abandoned.

SupSuper:
@stavrossk Rumors of our death have been greatly exaggerated. :)

@Player701 I can't believe people are still using MIDIs in 2018. :P In fact we strongly discourage them because they break the music volume slider. But if people really want...
Optimally we would just pick .MID first and this wouldn't be an issue. But as you said the digital versions include both, and while they are 99% the same, the GM.CAT version loops nicer. So them's the breaks.

Your PR looks fine, although I would rename the combobox options "GM" and "MIDI" for simplicity's sake:
1. We can't have hardcoded English strings, and I don't wanna add new strings for the translators and deal with UI issues.
2. They aren't very friendly, but this isn't an option intended for the average user anyways. Most people don't even know what "Adlib" is anymore.
3. And in all honesty, there's no reason anyone should have to pick the "GM" option. But trying to hide it from the combobox would just complicate the code unnecessarily.

Player701:
Renamed the music formats in the options to "GM" and "MIDI" accordingly.

Regarding MIDI files: I am well aware they are considered outdated, but I've been a strong fan of MIDI music (played back by the built-in Windows synthesizer, to boot) ever since I first played Doom. There are a lot of user-made add-ons for Doom and Doom II which provide custom MIDIs, and some of them fit X-COM's atmosphere really well. I've been using a custom MIDI pack with vanilla X-COM (Collector's Edition) for a really long time, and after switching to OpenXcom, I made a mod so that it wouldn't replace any original files. Basically, that's how I ran into the issue in question.

I've also heard that FluidSynth is a good choice for playing back MIDI files (it is used in GZDoom), though I currently do not have enough time to look into it. Besides, C++ is not my main programming language, so it'd take more time for me to integrate FluidSynth than for someone who knows C++ better. Perhaps you could look into it someday, if you are interested.

SupSuper:
SDL_Mixer supports Fluidsynth out of the box, however it's not enabled by default, so it's kind of a hassle to set up since you'd have to compile SDL_Mixer yourself. If you can get it running though, you should be able to use it without any code changes.

PR #1187, opened 2018-07-01
Allow mods to define whether a unit can be captured alive

Player701:
This pull request has been created to implement [feature request #1398](https://openxcom.org/bugs/openxcom/issues/1398) filed on the OpenXcom bug tracker.

The logic which determines whether a unit can be captured or not is currently [hard-coded](https://github.com/SupSuper/OpenXcom/blob/master/src/Battlescape/UnitDieBState.cpp#L214). The unit can be captured alive if, and only if, it does not explode on death. There is no way to make a unit not capturable without making it explode on death, nor is it possible to make an alien which explodes on death capturable.

The proposed change introduces a new unit property called "capturable", which can take one of the three possible values:

* 0 (CAP_NOT_CAPTURABLE) - the unit cannot be captured alive by any means (it will simply die when it falls unconscious).
* 1 (CAP_ALWAYS_CAPTURABLE) - the unit can be captured even if the original logic wouldn't have allowed it (i.e. the unit explodes on death)
* 2 (CAP_DEFAULT) - the default, previously hard-coded logic applies (see above)

The default value for all units is 2 (CAP_DEFAULT), so that compatibility with existing mods is not broken. The original rulesets are NOT changed.

The implementation has been tested to work correctly for all values of the new property (battle generator + debug mode).

### DIFF
--- a/src/Savegame/BattleUnit.cpp
+++ b/src/Savegame/BattleUnit.cpp
@@ -54,7 +54,7 @@ BattleUnit::BattleUnit(Soldier *soldier, int depth) :
 	_expBravery(0), _expReactions(0), _expFiring(0), _expThrowing(0), _expPsiSkill(0), _expPsiStrength(0), _expMelee(0),
 	_motionPoints(0), _kills(0), _hitByFire(false), _hitByAnything(false), _moraleRestored(0), _coverReserve(0), _charging(0), _turnsSinceSpotted(255),
 	_statistics(), _murdererId(0), _mindControllerID(0), _fatalShotSide(SIDE_FRONT), _fatalShotBodyPart(BODYPART_HEAD),
-	_geoscapeSoldier(soldier), _unitRules(0), _rankInt(0), _turretType(-1), _hidingForTurn(false), _respawn(false), _capturable(CAP_DEFAULT)
+	_geoscapeSoldier(soldier), _unitRules(0), _rankInt(0), _turretType(-1), _hidingForTurn(false), _respawn(false), _capturable(true)
 {
 	_name = soldier->getName(true);
 	_id = soldier->getId();
@@ -3291,18 +3291,9 @@ void BattleUnit::resetHitState()
 /**
  * Gets whether this unit can be captured alive (applies to aliens).
  */
-bool BattleUnit::isCapturable() const
+bool BattleUnit::getCapturable() const
 {
-	switch (_capturable)
-	{
-		case CAP_NOT_CAPTURABLE:
-			return false;
-		case CAP_ALWAYS_CAPTURABLE:
-			return true;
-		case CAP_DEFAULT:
-		default:
-			return !(getSpecialAbility() == SPECAB_EXPLODEONDEATH || getSpecialAbility() == SPECAB_BURN_AND_EXPLODE);
-	}
+	return _capturable;
 }
 
 }
--- a/src/Savegame/BattleUnit.cpp
+++ b/src/Savegame/BattleUnit.cpp
@@ -54,7 +54,7 @@ BattleUnit::BattleUnit(Soldier *soldier, int depth) :
 	_expBravery(0), _expReactions(0), _expFiring(0), _expThrowing(0), _expPsiSkill(0), _expPsiStrength(0), _expMelee(0),
 	_motionPoints(0), _kills(0), _hitByFire(false), _hitByAnything(false), _moraleRestored(0), _coverReserve(0), _charging(0), _turnsSinceSpotted(255),
 	_statistics(), _murdererId(0), _mindControllerID(0), _fatalShotSide(SIDE_FRONT), _fatalShotBodyPart(BODYPART_HEAD),
-	_geoscapeSoldier(soldier), _unitRules(0), _rankInt(0), _turretType(-1), _hidingForTurn(false), _respawn(false)
+	_geoscapeSoldier(soldier), _unitRules(0), _rankInt(0), _turretType(-1), _hidingForTurn(false), _respawn(false), _capturable(CAP_DEFAULT)
 {
 	_name = soldier->getName(true);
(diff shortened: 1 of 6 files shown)


SupSuper:
I would make the default option 0, since it's assumed the first option is the default and enums normally default to 0, and name it CAP_VANILLA since "default" says nothing about the behavior (possibly explain in a comment somewhere what the vanilla behavior actually is as well).

I'll let @Warboy1982 review the rest of it.

Player701:
The reasoning behind the values is that 1 and 0 intuitively correspond to "yes" and "no" ("true" and "false") while 2 is a fallback value and there is little point to use it in mods.

From a programmer's point of view, 0 should be the default, but then it will be hard to understand what 1 and 2 mean without extra documentation. This will make life a little more difficult for mod authors.

But I can of course change it if you really want to.

WalterBarrett:
Perhaps 0 (CAP_NEVER_CAPTURABLE), 1 (CAP_ALWAYS_CAPTURABLE), and 2 (CAP_UNLESS_EXPLODES)?

Player701:
@WalterBarrett: This looks fine to me, but we first need to decide whether or not 0 should be the default.

To everyone interested in the reasoning behind this whole PR: there is a link to a [bug report](https://openxcom.org/bugs/openxcom/issues/1263) on the feature request page on OXC bug tracker. Also giving additional explanation here:

It is definitely not possible in vanilla X-COM to capture a live Cyberdisc or Sectopod. This is confirmed not only by my own experience but also by these excerpts from the UFOpaedia website:

> The two robotic units (Sectopods and Cyberdiscs) cannot be captured alive. Therefore, only 27 of the race/rank combos can be kept in a containment facility. 

Source: https://www.ufopaedia.org/index.php/Alien_Containment#Unique_Alien_Types

> The two robotic units (Sectopods and Cyberdiscs) cannot be captured alive or studied directly; there isn't even a RESEARCH.DAT entry for them. To generate a live alien UFOpaedia entry on them, you must research alien Medics of any other species. 

Source: https://www.ufopaedia.org/index.php/Live_Alien_Research#Alien_Species

> Note: it is not possible to capture one, the only way to get this report is to get it from a captured medic

Source: https://www.ufopaedia.org/index.php/Sectopod#Live_specimen

Note that these two units, while not capturable, are also **NOT** immune to stun damage. When they fall "unconscious", they are counted as dead by the game.

OpenXcom stays true to vanilla regarding Cyberdiscs but not Sectopods. This is because the "unconscious equals dead" logic is tied to the Cyberdisc's self-exploding ability. Apparently, the dev team considers vanilla behavior a bug (even though everything points to the contrary because there isn't even _a research project_ for a Sectopod Terrorist in the game files). I wish it were possible to mod the original behavior back into the game, hence this PR. The currently hard-coded logic also prevents capturing live Bio-Drones in TFTD, despite it being allowed in vanilla (at least Collector's Edition). With this PR, a mod could be created that re-implements vanilla behavior for both EU and TFTD. Also, extra customization will be available to mod authors if they wish to add new self-exploding units into the game.

SupSuper:
After discussing with @Warboy1982 we're gonna be further investigating this to figure out the specifics of the vanilla behavior and where OpenXcom is in the right or wrong.

In any case I think this feature is misnamed, because it doesn't have anything to do with "capturing" (i.e. recovery), what actually happens is there's no "unconscious" status. A modder might wanna make a friendly unit be instantly killable, or an enemy unit go "unconscious" but not "capturable" (not count towards recovery).

Warboy1982:
yeah, i can't find anything that supports the assertion we've made here in our code, melee negates explodeOnDeath, but nothing negates stunning and going unconscious, apart from the "disabledInventory" flag, which as far as i can tell is only applicable to xcom tanks

Player701:
@SupSuper,  @Warboy1982  I understand if you have to look further into this stuff. In the end, I hope it will eventually be possible to reinstate the original behavior (Sectopods not capturable, Bio-Drones are) either as an option or via a mod.

Warboy1982:
further research reveals: vanilla arranged "accidents" for live captured units with no inherent research value, cyberdisks and sectopods had no entry in the research.dat file, and so they got flagged as "of no interest to the science team" and would be recovered as corpses.

SupSuper:
@Player701 Given @Warboy1982 's investigation, I think the correct course is:
- make `capturable` just a `true`/false flag. (default `true`)
- Cyberdisc and Sectopod Terrorists are set to `false` in the vanilla ruleset. (if anyone wants to bring them back they can just make a mod)

No point in keeping the "vanilla" option that wasn't there in the first place.

src/Savegame/BattleUnit.cpp
```diff
@@ -187,6 +187,7 @@ BattleUnit::BattleUnit(Unit *unit, UnitFaction faction, int id, Armor *armor, St
 	_spawnUnit = unit->getSpawnUnit();
 	_value = unit->getValue();
 	_faceDirection = -1;
+	_capturable = unit->getCapturable();
```
SupSuper: Should this be duplicated in BattleUnit? Could just be directly referenced from Unit.

Player701:
The property has been changed to a boolean flag per @SupSuper's request.

@SupSuper: I tried to follow the code style when making the PR, and I noticed that several of Unit's properties are duplicated in BattleUnit (like special ability, for example). I think there is a purpose behind it - I'm not sure if I can reference the Unit directly from the BattleUnit I have in UnitDieBState, since the pointer to it is initialized to 0 in some cases (when creating a BattleUnit from a Soldier, for example). IMO it is better to save the value in BattleUnit and use it from there without additional checks - less errors this way. Or do you mean something else by your question?

SupSuper:
@Player701 Fair enough, I guess either a null-check or an extra variable works out the same. Merged!

Player701:
@SupSuper Thanks a lot :)

PR #1189, opened 2018-07-25
Make CMake include the icon in the Mac bundle

dopplershift:
This is the magic incantation I had to use in order to get the .app bundle to show up with the appropriate icon on MacOS.


dopplershift:
All came from this [Stack Overflow post](https://stackoverflow.com/questions/27181879/cmake-copy-icon-to-bundle-resources).

SupSuper:
Hmm doesn't CPack cover it? https://github.com/SupSuper/OpenXcom/blob/master/cmake/modules/apple.cmake

dopplershift:
Without at least some of those lines, I don't see `AppIcon.icns` in `openxcom.app`. I can see if I can reduce the set though.

SupSuper:
Actually on further inspection, it makes sense, CMake just takes everything in as "source" and figures out what to do with it from the type. We had to do the same thing for the Windows icon, see here: https://github.com/dopplershift/OpenXcom/blob/9b59e921d1cb1b0f5e7904ee4ca4d51ea880d5a4/src/CMakeLists.txt#L6
Might be worth changing that to also use `$openxcom_icon` for consistency.

PR #1190, opened 2018-07-27
Cleaning up OSX build steps

rcreasey:
Removing the old xcodeproj and adding instructions to just let cmake generate the project file.

Tested by building in both Xcode and make modes.


dopplershift:
With this PR (and current master), I get:
```
ERROR: Failed to open directory: common
Make sure you installed OpenXcom correctly.
```

This was after doing a `git clean -fxd`, `cmake .` and `make -j 6` from the root of the repository. If I roll back one commit from master, it works fine.

Also, I'm surprised by the changes to `.gitignore`--isn't the point to ignore the build artifacts so that `git status` shows only files you may want to commit (and useful changes don't get lost)?

SupSuper:
I suppose the problem here is whatever fixes we did to make it work for Xcode broke it for other compilers. Optimally what should happen when building app bundles, regardless of compiler, is:
- OpenXcom bin folders (common, standard, etc) get copied to `Contents/Resources`.
- User can opt to copy their UFO/TFTD installs either to `Contents/Resources` or `~/Library/Application Support/OpenXcom`.
- OpenXcom should be able to locate the files in either location.

As for build artifacts, I agree they shouldn't be part of Git. Normally the standard way is to have out-of-source builds so you can just delete the build path when you're done: https://stackoverflow.com/questions/9680420/looking-for-a-cmake-clean-command-to-clear-up-cmake-output

rcreasey:
Hrm, you're right.  It does seem to be broken when trying to load the data path from the bundle:

From a fresh install on a new machine:
```
~/Library/Application Support/OpenXcom
❯ cat openxcom.log
[27-07-2018_13-04-32]	[INFO]	Data folder is:
[27-07-2018_13-04-32]	[INFO]	Data search is:
[27-07-2018_13-04-32]	[INFO]	- /Users/rcreasey/Library/Application Support/OpenXcom/
[27-07-2018_13-04-32]	[INFO]	- /Users/Shared/OpenXcom/
[27-07-2018_13-04-32]	[INFO]	- ./
[27-07-2018_13-04-32]	[INFO]	User folder is: /Users/rcreasey/Library/Application Support/OpenXcom/
[27-07-2018_13-04-32]	[INFO]	Config folder is: /Users/rcreasey/Library/Application Support/OpenXcom/
[27-07-2018_13-04-32]	[INFO]	Options loaded successfully.
[27-07-2018_13-04-32]	[INFO]	SDL initialized successfully.
[27-07-2018_13-04-32]	[INFO]	SDL_mixer initialized successfully.
[27-07-2018_13-04-32]	[INFO]	requested file not found: openxcom.png
[27-07-2018_13-04-32]	[INFO]	Attempting to set display to 640x400x8...
[27-07-2018_13-04-32]	[INFO]	Display set to 640x400x8.
[27-07-2018_13-04-32]	[INFO]	Loading data...
[27-07-2018_13-04-32]	[ERROR]	Failed to open directory: common
```

Since the Xcode artifact isn't copying data into it's bundle (like a nightly would), I'll see if I can get that fixed as well.

rcreasey:
How did data loading in Bundles ever work?
[src/Engine/CrossPlatform.cpp#L199-L226](https://github.com/SupSuper/OpenXcom/blob/master/src/Engine/CrossPlatform.cpp#L199-L226)

SupSuper:
OSX defaults the working directory `./` to `Contents/Resources` IIRC. There's no other way to get to it without getting into Obj-C calls.

dopplershift:
When it's been working, I noticed `current/` is created in a the root of the repository, which is the parent directory of `openxcom.app`.

SupSuper:
Ah I see what the problem is. We customized the `SDLmain.m` to set the working directory to `/Contents/Resources/`, so that OpenXcom could load files from the bundle. However this was only used in the XCode project.

Without it, SDL sets the working directory to the parent directory of the .app, so it happened to work for you @dopplershift because that's where the data was being (incorrectly) copied.

The correct course of action should be to restore `SDLMain.m` and point to it via MACOS_SDLMAIN_M_PATH: https://github.com/SupSuper/OpenXcom/blob/master/CMakeLists.txt#L36
(assuming it works)

adesst:
> With this PR (and current master), I get:

>ERROR: Failed to open directory: common
>Make sure you installed OpenXcom correctly.

I run debian8 64bit, had the same error.
Don't know if my post is out of topic
since this thread is OSX

I **solved** it by adding link of 
Openxcom/bin/common 
Openxcom/bin/standard
Openxcom/bin/UFO

into my ~/.local/share/openxcom/ directory
$ ls -l ~/.local/share/openxcom/
total 12
lrwxrwxrwx 1 adesanto adesanto   34 Jul 27 20:29 common -> /opt/cpp/games/OpenXcom/bin/common
drwxrwxr-x 2 adesanto adesanto 4096 Jul 27 20:27 mods
-rw-r--r-- 1 adesanto adesanto 2313 Jul 28 11:13 openxcom.log
lrwxrwxrwx 1 adesanto adesanto   36 Jul 27 20:30 standard -> /opt/cpp/games/OpenXcom/bin/standard
lrwxrwxrwx 1 adesanto adesanto   31 Jul 27 20:30 UFO -> /opt/cpp/games/OpenXcom/bin/UFO
drwxrwxr-x 2 adesanto adesanto 4096 Jul 28 09:11 xcom1

rcreasey:
Once this PR is accepted I'll also update the wiki with updated build steps for osx:

https://www.ufopaedia.org/index.php?title=Compiling_(OpenXcom)
https://www.ufopaedia.org/index.php/Compiling_with_Xcode_(OpenXcom)
https://www.ufopaedia.org/index.php/Compiling_with_brew_(OpenXcom)

dopplershift:
Works for me with `cmake .` and `make -j 6`. `common/` and friends are in `Contents/Resources/`.

adesst:
i git-pull , make and remove all linked files of common, standard and UFO
in my ~/.local/share/openxcom

Start the game, and the missing **common/ error** is raised.

i re-create the link files into ~/.local/share/openxcom **manually** and the game works again.

rcreasey:
Wait. If you're building this on Linux, this PR only changes Mac bundling and Linux is unaffected. 

Linux binaries aren't bundles and wouldn't automatically move then out of bin into your data path. Putting your data files directly into the shared system location is the preferred method on Linux AFAIK.

PR #1193, opened 2018-08-01
Update ax_check_gl from autoconf-archive

scarabeusiv:
Note one tiny thing might be sensible to run something along following lines in .travis.yml but since you already have quite large matrix I am not sure if you wanna do it:

```bash
sh autogen.sh
mkdir build
pushd build
../configure
make distcheck
popd
```

The make distcheck generates release-able tarball and ensures that such tarball is still installable and uninstallable without any issues.


SupSuper:
Yeah our matrix already takes a long time, if I'm gonna add any other toolchains in there I'm gonna drop something else, like maybe stick to only Release configuration.

PR #1194, opened 2018-08-11
Fixed BGM and depth for the GRUNGE terrain

Player701:
This PR is the proposed fix for [issue #1415](https://openxcom.org/bugs/openxcom/issues/1415) filed on the OpenXcom bug tracker.

In short: GRUNGE terrain seems to have wrong background music set in the standard ruleset (GMTACDRY instead of GMTACWET). It also doesn't have any depth set, which defaults to a depth of 0, though this is overridden by the alienDeployments.rul entry for the corresponding mission type (artifact site level 2). Nevertheless, I thought it would not hurt to specify the depth in this file also, considering it is specified for every other terrain where it differs from the default values, even if those terrains have it overridden during initialization.

BTW, considering the depth override, there seem to be certain inconsistencies between alienDeployments.rul and terrains.rul:

- Alien Colony level 1 deployment sets the depth range to [3, 3] but ENTRY terrain has a range of [2, 3]
- Alien Colony level 2 deployment sets the depth range to  [3, 3] but A_BASE terrain has a range of [1, 1]
- Artefact Site level 1 deployment sets the depth range to [1, 3] but ALART terrain has a range of [1, 2]. However, the override here doesn't seem to cause any effect as it uses the "siteDepth" parameter instead of just "depth". I couldn't find any references to this parameter within the game code outside of the AlienDeployment class; it seems to be unused by the game.

Which values are correct?

### DIFF
--- a/bin/standard/xcom2/terrains.rul
+++ b/bin/standard/xcom2/terrains.rul
@@ -849,7 +849,8 @@ terrains: #done
         groups: 10
   - name: GRUNGE
     music:
-      - GMTACDRY
+      - GMTACWET
+    depth: [3, 3]
     ambience: 67
     ambientVolume: 0.25
     mapDataSets:


Player701:
See here: https://github.com/SupSuper/OpenXcom/pull/1195

Player701:
Thanks! Maybe you could check this part also:

> BTW, considering the depth override, there seem to be certain inconsistencies between alienDeployments.rul and terrains.rul:
> - Alien Colony level 1 deployment sets the depth range to [3, 3] but ENTRY terrain has a range of [2, 3]
> - Alien Colony level 2 deployment sets the depth range to [3, 3] but A_BASE terrain has a range of [1, 1]
> - Artefact Site level 1 deployment sets the depth range to [1, 3] but ALART terrain has a range of [1, 2]. However, the override here doesn't seem to cause any effect as it uses the "siteDepth" parameter instead of just "depth". I couldn't find any references to this parameter within the game code outside of the AlienDeployment class; it seems to be unused by the game.

I'm not sure which values are valid, as I do not have access to decompiled sources of the original game nor any experience in decompilation / disassembly.

Warboy1982:
siteDepth is used to determine if a craft can actually land at the site. vanilla craft can land at all depths but even the original TFTD had code that stopped craft from landing if the siteDepth exceeded the allowable depth of a given craft. this is/was vestigial, but something left in for modders to use (for example, a skyranger shouldn't be able to land at an underwater site)

Player701:
Pardon me, but searching for "siteDepth" in the source code yields results ONLY in the following files:
- AlienDeployment.cpp
- AlienDeployment.h

Specifically, [the part where minSiteDepth and maxSiteDepth values are extracted from the save file](https://github.com/SupSuper/OpenXcom/blob/master/src/Mod/AlienDeployment.cpp#L165), as well as [the getter methods](https://github.com/SupSuper/OpenXcom/blob/master/src/Mod/AlienDeployment.cpp#L420) for these values. However, there is no mention of the latter anywhere else in the code, so these methods are never called during the execution. Unless I'm missing something way more obvious than I'm looking for, of course... but there doesn't seem to be any additional code in those files which deals with these values.

SupSuper:
I removed the code as it was being used incorrectly (checking for mission depth on dogfights): https://github.com/SupSuper/OpenXcom/commit/d5d3127e49203004463f6c89d0720b1cd3f6f7b1

If a modder actually intends to use it, it would have to be rewritten. As it is, it might be better to remove the unused code and values to avoid confusion.

Player701:
...and therefore, the question about the depth values in alienDeployments.rul and possible typo in Artifact Site stage one deployment entry ("siteDepth" vs "depth") still stands. Like I said, I'm not sure which values are correct (those in alienDeployments.rul or terrains.rul) because I don't have access to the original code in any human-readable form.

SupSuper:
`siteDepth` was only intended for the Geoscape and is completely unused as far as I can tell. This is completely separate from the `depth` value in the terrains is what's used in the Battlescape and should be correct.

PR #1195, opened 2018-08-12
Fix depth and music not changing between stages in multi-stage missions

Player701:
This PR fixes [issue #1416](https://openxcom.org/bugs/openxcom/issues/1416) filed on the OpenXcom bug tracker.

In short: depth and music will now change between mission stages as they should.

In detail: this PR adds two new private methods to BattlescapeGenerator.cpp: setDepth and setMusic with code refactored into them from the run method. These methods are now called in the run method as well as the nextStage method in BattlescapeGenerator.cpp. Note that these methods have to know whether the mission is currently at the first stage or progressing to a new stage.

Notes about the setDepth method:
- First we need to check if the depth is already set (not 0) by the battle generator (original behavior). This also requires the mission to be at the first stage.
- We assume that an AlienDeployment cannot force an underwater terrain to have a depth of 0 (land), so 0 instead means "use terrain depth", and thus the check for 0 here does not need any extra checks.
- Note that with the current code it is possible for an AlienDeployment to override a land terrain (depth 0) to be an underwater terrain. This is possibly unintended, but it is the current behavior (not changed by this PR). To rectify this, it would be necessary to also check for _terrain->getMaxDepth() > 0 when checking the AlienDeployment depth.
- If the terrain has a depth of 0, it needs to be set at a new stage only, because otherwise it would already be 0.

Notes about the setMusic method:
- We assume that if an AlienDeployment has an empty music vector, this means "use terrain music", so the check for emptiness here does not need any extra checks.
- If a terrain does not have any music set, we will set an empty music at a new stage only, because otherwise it would already be empty (note that empty music name seems to default to GMTACTIC).

### DIFF
--- a/src/Battlescape/BattlescapeGenerator.cpp
+++ b/src/Battlescape/BattlescapeGenerator.cpp
@@ -371,6 +371,7 @@ void BattlescapeGenerator::nextStage()
 	ruleDeploy->getDimensions(&_mapsize_x, &_mapsize_y, &_mapsize_z);
 	size_t pick = RNG::generate(0, ruleDeploy->getTerrains().size() -1);
 	_terrain = _game->getMod()->getTerrain(ruleDeploy->getTerrains().at(pick), true);
+	setDepth(ruleDeploy, true);
 	_worldShade = ruleDeploy->getShade();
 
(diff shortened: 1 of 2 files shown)


Player701:
See here: https://github.com/SupSuper/OpenXcom/pull/1195

PR #1197, opened 2018-08-14
Display the number of alien bases destroyed and X-Com bases lost in statistics

Player701:
The number of alien bases destroyed and the number of X-Com bases lost are calculated in StatisticsState.cpp but not displayed in the statistics window. This PR makes these numbers display in the statistics window as they were probably intended to.


Player701:
Huh? Well, at least it's nice to see this project has not been abandoned, even though I have no idea what was wrong with this PR.

SupSuper:
It's redundant since that stuff is already in the diaries. The only reason it's in the code is to calculate other stats.

Player701:
It was missing from the post-endgame statistics window, that's what this PR was about. If I remember right, the corresponding variables were not used anywhere else. But I guess some code may have been changed over all this time, or I might be misremembering something.

PR #1198, opened 2018-09-21
Upgrade to SDL2

JayDT:
Upgraded to SDL2 and appended required dependencies and a few bugfixes.


SupSuper:
This is just a mess:
- I'm not reviewing ~2 million LOC changes.
- There's already a SDL2 branch: https://github.com/sfalexrog/OpenXcom/tree/sdl2
- Dependencies don't belong in the repository, use packages or submodules.
- Main repository can't switch to SDL2 as some platforms still depend on SDL1.2 (at most we will dual-support both).
- Various other junk files and unrelated changes in one PR.

However if you can PR the bugfixes separately, that'd be appreciated.

PR #1200, opened 2018-10-11
new gameloop and more data in the fpscounter

StoddardOXC:
This is a bit of rework of the game loop and the FPS counter.

Game loop changes ensure that input events are polled only once per frame.
It then draws the frame and sleeps the rest of the alotted time (fps limit) in one go.
This reportedly results in noticeable decrease of CPU consumption especially when the game is in background.

The FPS counter is changed to display timings of all of the stages - input, logic, blit and idle, and the resulting FPS.

I had to add another hardcoded font - a 4x6 BSD/CCBY3 licensed one. This would be useful for any other debugging info display.


src/Engine/Game.cpp
```diff
@@ -508,18 +525,20 @@ void Game::setSavedGame(SavedGame *save)
  */
 Mod *Game::getMod() const
 {
-	return _mod;
+	return  _mod_ready ? _mod : NULL;
```
Yankes: How it will help? Isn't this code call from other thread? Per standard game can read `true` in other thread.

src/Engine/Game.cpp
```diff
@@ -257,56 +261,69 @@ void Game::run()
 					break;
 			}
 		}
-		
-		// Process rendering
-		if (runningState != PAUSED)
+		Uint32 frameNominalDuration = 1; // 1e3 FPS should be enough for anybody
+		if (Options::FPS > 0 )
```
SupSuper: This should be `fpsLimit > 0`, or OpenXcom will crash when Options::FPSInactive = 0

SupSuper:
Please, no C++11. :)

I can't really comment much on the game loop stuff. I'm wary of touching anything that internal, but it looks good in theory.

I've got more feedback about the FPS Counter changes. Players use it too, and all the extra info is just noise they won't understand. Plus it's got some issues:
- Currently doesn't refresh outside of state changes. If I add `redraw = true` to addFrame then it updates too frequently, becoming an eyesore.
- There's no point in hardcoding a font unless it's used before loading resources (like the terminal font). It should be added to `Font.dat` per norm to avoid polluting code.
- Plus adding a whole font just for the FPS Counter seems excessive. If you want a small font you can use the GEO fonts.

src/Interface/FpsCounter.cpp
```diff
@@ -88,37 +93,35 @@ void FpsCounter::handle(Action *action)
...
 	Surface::draw();
+	std::ostringstream actual_text;
+	actual_text<<"H:"<<std::setw(2)<<static_cast<int>(round(_input.get()));
+	actual_text<<" L:"<<std::setw(2)<<static_cast<int>(round(_logic.get()));
+	actual_text<<" B:"<<std::setw(2)<<static_cast<int>(round(_blit.get()));
+	actual_text<<" I:"<<std::setw(2)<<static_cast<int>(round(_idle.get()));
+	actual_text<<" F:"<<std::setw(2)<<static_cast<int>(round(_frame.get()));
+	actual_text<<"/"<<std::setw(2)<<static_cast<int>(_limit);
+	actual_text<<" FPS: "<<std::setw(2)<<static_cast<int>(round(1000.0/_frame.get()));
+	_text->setText(Language::utf8ToWstr(actual_text.str().c_str()));
```
SupSuper: If you use `wostringstream` you don't need this conversion.
Also, some spacing wouldn't hurt.

SupSuper:
> Why? No, seriously, I sincerely don't understand.
> Is there a secret compiler version baseline?
> 
We're keeping it to C++03 and will re-evaluate after 2.0. It's hard to justify kicking everyone off their compilers when OpenXcom has remained largely the same (same reason SDL2 is still up in the air).

> I hope you at least agree that the EMA filter and keeping tabs on all phases that the Game instance has access to is worth it. If people like the one-number fps counter, nothing prevents having another internal metrics display covering half a screen (half of a reason for embedding a full-ascii font).
> 
I'll have to take your word for it as all I see are values flickering between 0 and 1.

> 
> Very strange. I tested in linux and win7, nothing like that. Or maybe I don't understand what you mean by 'state changes'.
> 
You never trigger a redraw, so `FpsCounter::draw` will never be called and the values on the screen will never change. Internally they are changing, they're just not being drawn.
Adding `_redraw = true` to `FpsCounter::addFrame` fixes it, but since this runs every frame (rather than every second), it's very visually distracting compared to the old FPS counter.

> 
> Using regular small font was the first approach, but I found no way to choose good enough color to setColor() the text both in geoscape and battlescape even in vanilla -- without making the main loop aware which state the game is in, not to mention various mod-introduced palette changes.
> 
> The vanilla-ish fonts require a certain setup of the palette, which is harder to achieve compared to a hardcoded 1-bit font.
> 
> The choice between using NumberText, which was lacking letters, and extending it to be a font in its own right (given that the font in question is <1Kb) was obvious.

We already have a 1-bit font, FONT_GEO_SMALL (in fact I could just replace NumberText with that). Here's how it looks hooked to that instead (I added the missing ASCII): https://i.imgur.com/1rqGT5n.gifv

I'll give you that some characters don't look as nice, but it's preferable to hardcoded content we can't ever edit anymore (that has historically bit us in the ass).

PR #1201, opened 2018-10-28
Silence cmake policy warnings on openGL

R1dO:
Hopefully without breaking other peoples builds


R1dO:
After a good discussion with SupSuper we came to the conclusion there is a better way for solving this.

Hence closing this PR

PR #1204, opened 2018-10-28
Cmake (opengl) and travis fixes

R1dO:


knapsu:
Commit `5f5ff124ec540138f381cdbfdb234b021c739102` replaces `${OPENGL_gl_LIBRARY}` with `OpenGL::GL`. This does require recent version of CMake. Ubuntu Xenial and Trusty have only CMake 3.5 in their repositories which is too old and will break the build.

Travis CI extends their Ubuntu Trusty images with CMake 3.9 so it works (`x86_64` arch).

For building AppImages Ubuntu Docker images are being used in the target arch.
Because there are no custom PPA repositories to get newer `cmake` package I've added additional build step to download precompiled version from https://cmake.org/ . This is ok for 64-bit build environment (using version 3.12)  but not for 32-bit as version 3.6 was the last released for `i386` arch.

Because of this x86 AppImage builds are currently failing.
https://travis-ci.org/knapsu/OpenXcom-AppImage/builds/449638183

Building cmake on its own sounds like a possible workaround but I very much don't like this idea.
Upgrading base buld system for Ubuntu Bionic is another solution. but this will raise AppImage dependencies. Probably a bad idea.

R1dO:
That is a whoopsy from my side [1].
Judging from the release notes it has been introduced in 3.8.

Anyhows, you can try replacing `OpenGL::GL` with `${OPENGL_LIBRARIES}` to see if that helps with your builds. That way we will not 'force' you to upgrade the ubuntu distribution (after all Trusty is still LTS untill april 2019) while it would still allow `GLVND` to link if a user wants to build against that. 
If it works please let us know so we can fix it.

Again my apologies for the oversight

[1] To my defense there was no indication on a required cmake version in that (and related) part of the  documentation.

PR #1205, opened 2018-10-28
Test speedup yaml osx

R1dO:
If this works, it will be deleted and merged into another PR


R1dO:
Didn't know i could test travis from my own account.

Sorry for the noise

PR #1206, opened 2018-10-29
Reduce osx build time

R1dO:
By using a Brewfile travis skips the `brew update` part which contributes significantly to the build time.
Contrary to the alternative (using homebrew packages inside `.travis.yml` a Brewfile allows to specify formulae options, needed for flac.

`yaml-cpp` no longer uses the `--static-lib` option (saves ~2 minutes).

The options: `--with-libmikmod --with-libvorbis` are part of he default formulae for `sdl_mixer` hence they are omitted.

Brewfile package information is available at: https://formulae.brew.sh/formula/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,12 +17,13 @@ addons:
       - libsdl-mixer1.2-dev
       - libsdl-image1.2-dev
       - libsdl-gfx1.2-dev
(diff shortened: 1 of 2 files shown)


rcreasey:
LGTM. 👍

PR #1207, opened 2018-11-03
Travis ccache

R1dO:
By using ccache we can reduce travis build time to a few minutes

Depending on the amount of source file changes


rcreasey:
This looks good to me as well, but it would be good to try and burn down the clang warnings

R1dO:
From test preformed the only option is see is to silence those warnings for yaml-cpp (they do not happen for openxcom builds) using:
`-DCMAKE_CXX_FLAGS="-Wno-unused-command-line-argument"`

Unfortunately `ccache` with `clang` does not work as nice as with `gcc` on travis.
And it seems yaml-cpp has no Cmakefunctionality build in we can use to say: hey we want to use ccache.

If that is acceptable i can add such a commit to this PR

PR #1209, opened 2018-11-04
Fix 32bit appimages not being able to build

R1dO:
Those builds use a system with an older CMake compiler. One that does not recognize imported targets.

This fix should still allow for both the legacy and GLVND openGL libraries.

Sorry for breaking your builds @knapsu


knapsu:
Hi @R1dO and thanks for the fix.

I was able to do a workaround by using Ubuntu Bionic image when building
https://github.com/knapsu/OpenXcom-AppImage/commit/1f92aad23d550991172b5f474ef19477e3ec791a
but it makes AppImage to run only on Bionic or later Linux installs.

Changing back to Trusty image in that case. Thanks!

R1dO:
Well you might have to wait till the PR is actually in before reverting back
(just noticed the commit on your repo ;))

PR #1213, opened 2018-12-04
Added user-friendly compilation script

Beep6581:
Hey

The script compiles OpenXcom on any distro. It is up to the user to
ensure the dependencies are in place.
To automatically clone/update git and compile:
    ./scripts/build-openxcom

I maintain similar compilation scripts for several projects to make compilation a breeze for anyone, without having to delve into often-outdated docs.


AlexMooney:
For what it's worth, this worked great on my Arch Linux system.  Thanks!

PR #1218, opened 2019-02-10
Use emplace instead of push

ShFil119:
It allows to construct in place.


MeridianOXC:
Why not asking if the devs are interested in such a change first? Emplace_back is c++11, which is currently unwanted in OpenXcom and your PR will most likely be rejected...

ShFil119:
CMake sets c++11 and I haven't seen anywhere information to avoid c++11.

PR #1236, opened 2020-01-22
CMake: Fix library name for yaml-cpp on macos

JonnyH:
The "YAML" hint ends up finding the C libyaml on my system, instead of
the intended libyaml-cpp.


JonnyH:
Without this this fails to build for me (using the homebrew libyaml-cpp on macos 10.15.2)

No idea how this worked previously? Maybe it's just that the build environment used the pkgconfig setup, or had a local "DEPS_DIR"?

So abusing the CI to see if this is harmless in whatever config that uses.

PR #1238, opened 2020-03-02
FLC player now uses standard compliant behavior on default

memmaker:
It only reverts back to a work-around for the TFTD video files

### DIFF
--- a/src/Engine/FlcPlayer.cpp
+++ b/src/Engine/FlcPlayer.cpp
@@ -350,9 +350,13 @@ void FlcPlayer::decodeVideo(bool skipLastFrame)
 			{
 				delay = _delayOverride > 0 ? _delayOverride : _headerSpeed * (1000.0 / 70.0);
 			}
-			else
+			else if (_useInternalAudio && !_frameCallBack) // this means TFTD videos are playing
 			{
 				delay = _videoDelay;
+			}
+			else
+			{
+				delay = _headerSpeed;
 			}
 
 			waitForNextFrame(delay);


src/Engine/FlcPlayer.cpp
```diff
@@ -99,6 +99,12 @@ FlcPlayer::~FlcPlayer()
  */
 bool FlcPlayer::init(const char *filename, void(*frameCallBack)(), Game *game, bool useInternalAudio, int dx, int dy)
 {
+	std::string tftdPrefix = "TFTD/ANIMS/";
+	if (strncmp(filename, tftdPrefix.c_str(), tftdPrefix.length()) == 0)
+	{
+		// activate special TFTD behaviour for the delay between two frames
+		_delayDependsOnAudio = true;
+	}
```
SupSuper: I don't think you need to add another flag as `_useInternalAudio` is only used for TFTD videos.

memmaker: Are you saying, that _useInternalAudio has basically the same semantics as the one I introduced, so I could use that one instead? Will look into that.

memmaker: _useInternalAudio seems to also be true when running the vanilla UFO intro sequence, not just on TFTD.

Warboy1982: the UFO intro also has "useUfoAudioSequence" to delineate it, since it pulls all the audio from the .CAT files rather than the .VID

memmaker: So I could use this check then: if (_useInternalAudio && !frameCallBack)?, right

PR #1239, opened 2020-03-15
Allow hi-res FLC video files to be played

memmaker:


memmaker:
Reasoning: The built-in FLC player supports both the older FLI and the newer FLC format. The latter supports resolutions up to 1280x1024.
It just bugs me to let perfectly functional features lie dormant. And the resolution already allows more than 320x200 so there is no reason to not fix the scaling.

PR #1245, opened 2020-04-06
Fixed average monthly rating calculation in StatisticsState

Player701:
This PR fixes [bug 1409](https://openxcom.org/bugs/openxcom/issues/1409) reported on the old bug tracker.

**Synopsis:** The value for "Average monthly rating" in the end-game statistics window is calculated incorrectly. It is derived only from the research scores vector, disregarding both XCOM and alien activity in regions. The proposed fix is to employ the same calculation as the one used in the geoscape graphs window: research score plus XCOM activity minus alien activity.

Do note that there are at least two other problems with this calculation:

- Only the last 12 months (at most) are used to calculate the average value. This is because the game is programmed to store at most 12 values of statistical data in each vector (I first mentioned this [on the forum](https://openxcom.org/forum/index.php?topic=6511.0) as issue no. 4). This also affects the values for total income and total expenditure. I can report this as a separate issue if necessary.
- The data for the current month is included, but we do not account for the fact that the month is not yet over, and we still consider it a full month rather than a fraction of a month. I haven't changed it because this PR is intended to address only the semantics of the score calculation. If necessary, I can make another change, or a separate PR.

### DIFF
--- a/src/Menu/StatisticsState.cpp
+++ b/src/Menu/StatisticsState.cpp
@@ -39,6 +39,7 @@
 #include "../Savegame/SoldierDeath.h"
 #include "../Savegame/BattleUnitStatistics.h"
 #include "../Savegame/Country.h"
+#include "../Savegame/Region.h"
 #include "../Savegame/AlienBase.h"
 
 namespace OpenXcom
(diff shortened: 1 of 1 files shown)


Warboy1982:
just pointing out... you forgot to add the 400 point bonus for each month after the first

Player701:
Does the game not already add it to the research scores vector at each monthly report?

https://github.com/OpenXcom/OpenXcom/blob/430bf7cf47193ed8f104466562b105c14a05355e/src/Geoscape/MonthlyReportState.cpp#L351-L353

At least, that's what I seem to have in researchScores in my save file, not having done any research for the last 3 months. The value for the last month is 0, though, seeing as it's not yet over. Do you mean that I should account for the last month's bonus?

PR #1253, opened 2020-04-18
Allow player access to the built-in melee attack for Lobstermen

### LINKED ISSUE #1252: No player access to Lobsterman's melee weapon
**Synopsis:** The Lobsterman race uses a built-in melee weapon (claws) in addition to any normal weapons they usually carry. While the AI has no problem using the melee weapon on player units, it does not seem to be possible for the player to use this weapon when a Lobsterman is under their control. This goes against the usual OpenXcom practice to allow access to built-in melee attacks of all alien units.

Note that the weapon in question does seem to have a graphic defined for it (provided by OpenXcom in [standard/xcom2/Resources/Weapons/Terror.png](https://github.com/OpenXcom/OpenXcom/blob/10b140d5592d48889b3bad3d105994364f33d3fa/bin/standard/xcom2/Resources/Weapons/Terror.png), 5th picture from the left), so this problem could either be a bug or a planned feature that was never implemented properly, considering that the Lobsterman is a kind of a special case, having both the ability to carry normal weapons and a built-in melee attack. Terror units that have both a ranged and a melee weapon (like the Bio-Drone and the Triscene) are handled correctly in this regard.

**Reproduction steps:** Same as in #1249, this is easily witnessed in debug mode, or with MC.

Player701:
This PR resolves #1252 by allowing access to the built-in melee weapon of a unit from the battlescape UI.

**Synopsis:** See #1252 for details.

**TL;DR:** Lobstemen melee attack cannot be accessed when the unit is under player control. I have attempted to rectify the problem by allowing access to the special melee weapon through either of the two hand slots, provided that at least one of them is unoccupied. If both hand slots are empty, the right hand slot is used by default.

Note that all changes are UI only. Also, the active hand is only changed if the weapon being used is from the inventory, preserving the original behavior.

I have tested this in various scenarios and haven't encountered any unusual behavior so far. But it'd still be best if someone else could review the changes - there may be some special case that I've missed.

### DIFF
--- a/src/Battlescape/BattlescapeState.cpp
+++ b/src/Battlescape/BattlescapeState.cpp
@@ -1113,10 +1113,16 @@ void BattlescapeState::btnLeftHandItemClick(Action *)
 
 		_battleGame->cancelCurrentAction();
 
-		_save->getSelectedUnit()->setActiveHand("STR_LEFT_HAND");
+		BattleUnit *unit = _save->getSelectedUnit();
+		BattleItem *leftHandItem = getLeftHandItem(unit);
+
+		if (leftHandItem != getSpecialMeleeWeapon(unit))
+		{
+			unit->setActiveHand("STR_LEFT_HAND");
+		}
+
 		_map->cacheUnits();
 		_map->draw();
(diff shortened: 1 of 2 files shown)


Player701:
Wow, thanks! For one fleeting moment, I thought this day would never come...

PR #1257, opened 2020-06-03
Fix edge case for items and live alien recovery after a mission.

R1dO:
Before this commit the game would not force the player to sell/destroy excess items/aliens after a mission under the following conditions:
* Too little items in base storage to restock the craft
* Alien containment has not enough free slots to hold all the captured live aliens
* User activated the option "storageLimitsEnforced"

After this commit the game behaves the same as you would normally see when there is enough to restock the craft.

--- Extra ---
Bug was introduced by c74a3522. 
By moving the `_noContainment` check the else clause for `_manageContainment` accidently became part of the `!_missingItems.empty()` control flow.

Just another testament that we are mere humans and not alien overlords ;)

### DIFF
--- a/src/Battlescape/DebriefingState.cpp
+++ b/src/Battlescape/DebriefingState.cpp
@@ -700,7 +700,7 @@ void DebriefingState::btnOkClick(Action *)
 			{
 				_game->pushState(new CannotReequipState(_missingItems));
 			}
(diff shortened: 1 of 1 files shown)


SupSuper:
Nice catch, thanks!

PR #1259, opened 2020-07-02
Fix all remaining compile warnings 

myk002:
Tested on Clang 10.0.0 and GCC 9.3.0 on Gentoo Linux.

This pull request fixes all remaining compile-time warnings and allows the code to compile cleanly.

- warnings about GL_GLEXT_VERSION are fixed by including third party includes with -isystem instead of -I (which directs the compiler to not generate warnings for those includes)
- ISO C++ forbids converting a string constant to ‘char*’ warnings are fixed by moving string constants into file static const char* vars and casting upon usage
- warnings about missing override keywords are fixed by adding the override keyword where appropriate


uqs:
Are you sure all places have been fixed? On FreeBSD with clang 10.0.1 I get this:

```
  CXX      src/Basescape/openxcom-BaseInfoState.o
In file included from src/Basescape/BaseInfoState.cpp:30:
src/Basescape/../Interface/TextEdit.h:59:7: error: 'OpenXcom::TextEdit::setFocus' hides overloaded virtual function [-Werror,-Woverloaded-virtual]
        void setFocus(bool focus, bool modal = true);
             ^
src/Basescape/../Interface/../Engine/InteractiveSurface.h:65:15: note: hidden overloaded virtual function 'OpenXcom::InteractiveSurface::setFocus' declared here: different number of parameters (1 vs 2)
        virtual void setFocus(bool focus);
                     ^
1 error generated.
```

myk002:
This PR addressed all compile warnings for Clang 10.0.0 and GCC 9.3.0 on Gentoo Linux. Other compiler versions on other operating systems may very well require further work.

PR #1268, opened 2020-09-12
Fix compilation in C++17 mode, where std::random_shuffle has been

uqs:
removed.

It's been deprecated as of C++11, so switch to std::shuffle() starting
with that.

### DIFF
--- a/src/Engine/RNG.h
+++ b/src/Engine/RNG.h
@@ -18,6 +18,7 @@
  * along with OpenXcom.  If not, see <http://www.gnu.org/licenses/>.
  */
 #include <algorithm>
+#include <random>
 #define __STDC_LIMIT_MACROS
 #include <stdint.h>
 
@@ -55,7 +56,11 @@ namespace RNG
 	template <typename T>
 	void shuffle(T &list)
 	{
+#if __cplusplus >= 201103L
+		std::shuffle(list.begin(), list.end(), std::random_device());
+#else
 		std::random_shuffle(list.begin(), list.end(), generateEx);
+#endif
 	}
 }
 


SupSuper:
This would ignore our RNG and not use the savegame seed. I assume to fix it "properly" you would have to convert our RNG to a std generator.

uqs:
Yes, so I guess it would allow people to do "save scumming", but I think that's better than not compiling at all :) 

If I wanted to turn the existing RNG into a generator, where would I instantiate that class? If you can point me there, that might get things started.

MeridianOXC:
> If I wanted to turn the existing RNG into a generator, where would I instantiate that class? If you can point me there, that might get things started.

FWIW, this is how it has been done in OXCE: https://github.com/MeridianOXC/OpenXcom/commit/09dd47a2d172dc3476812310011513e81cd0483e

uqs:
Perfect, thanks!

@SupSuper, as I don't want to claim credit for those changes, could you please push that into your repo?

```
% git remote add meridian git@github.com:MeridianOXC/OpenXcom.git
% git fetch --all
% git cherry-pick 7d95bd3578e8fb73b5f00d24349317e8c2a54982
% git cherry-pick 77c9464ac6594d326f86b59dee35330dc7c47dfd
% git cherry-pick 09dd47a2d172dc3476812310011513e81cd0483e
<test stuff>
% git push
```

uqs:
This problem is still open. The 3 commits from meridian's repo would need to be cherry picked to make the RNG work in C++17 mode.

MeridianOXC:
Here's a consolidated version: https://github.com/MeridianOXC/OpenXcom/commit/c27855cd0f879b71d5825485f8df2d588ffc39d5

but I have no idea how to test if it compiles in c++03
I just don't have any such compiler
(I removed all c++14 code and put all c++11 code within #ifs)

MeridianOXC:
Update:
MacOS build failed... it needs the `constexpr` that I removed.
Adding it back helped: https://github.com/MeridianOXC/OpenXcom/commit/69b9071cec11ea6eee931c971517cfb833685bdf
But it will also need to be put inside #if for backwards compatibility.
All in all, a major pain the ass...

MeridianOXC:
The issue has been fixed by removing usage of std::shuffle and std::random_shuffle completely.

PR #1271, opened 2020-09-22
Fix build with clang 10.0.1 on FreeBSD

uqs:
This is needed in addition to "CXXFLAGS=-O2 -g -std=c++11" and disabling
-Werror (due to unused functions, clang seems to be rather strict here).

Also, the setFocus method wasn't actually overridden, but hidden, which
makes clang warn and given the default of -Werror also error out.

This can probably be done in a smarter way ...

### DIFF
--- a/src/Interface/TextEdit.h
+++ b/src/Interface/TextEdit.h
@@ -54,15 +54,17 @@ public:
 	/// Cleans up the text edit.
 	~TextEdit();
 	/// Handle focus.
-	void handle(Action *action, State *state);
+	void handle(Action *action, State *state) override;
 	/// Sets focus on this text edit.
-	void setFocus(bool focus, bool modal = true);
+	void setFocus(bool focus, bool modal);
+	// Override the base class' method properly.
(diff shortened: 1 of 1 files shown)


src/Interface/TextEdit.h
```diff
@@ -54,15 +54,17 @@ class TextEdit : public InteractiveSurface
 	/// Cleans up the text edit.
 	~TextEdit();
 	/// Handle focus.
-	void handle(Action *action, State *state);
+	void handle(Action *action, State *state) override;
 	/// Sets focus on this text edit.
-	void setFocus(bool focus, bool modal = true);
+	void setFocus(bool focus, bool modal);
+	// Override the base class' method properly.
+	void setFocus(bool focus) override { setFocus(focus, false); };
```
SupSuper, line 61: Shouldn't this be `setFocus(focus, true);` ?

uqs: Oops, fixed.

PR #1279, opened 2020-11-04
Fix GH actions

uqs:
This fixes the MacOS build on GH (Travis unaffected) and removes some misleading and non-nonsensical combinations of OS+compiler.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,8 +28,6 @@ jobs:
               libyaml-cpp-dev
             compiler: gcc
             cpp-compiler: g++
-            cflags: "-Wall -Wextra"
-            cxxflags: "-Wall -Wextra"
 
           - name: Linux
             os: ubuntu-20.04  # -latest is still stuck at 18.04
@@ -39,8 +37,6 @@ jobs:
               libyaml-cpp-dev
             compiler: clang
             cpp-compiler: clang++
-            cflags: "-Wall -Wextra"
-            cxxflags: "-Wall -Wextra"
 
           - name: Windows
             os: windows-latest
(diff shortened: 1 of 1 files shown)


SupSuper:
What are the consequences of storing artifacts in GitHub? I don't wanna break their free generosity by overflowing any limits.

We have our own server for making the builds on the site, so it's a lot faster than these options as there's no constant "reinstall everything from scratch" required. But of course it's impossible to do that with OSX, so that would be a nice bonus.

uqs:
There is a retention limit for these sorts of build artifacts.  The documentation doesn't actually say what the default is, but this commit here https://github.com/actions/upload-artifact/commit/59018c2f85dd0e101b75544aa87f13bb0c94e0b7 leads me to believe the max is 90d and the default can be changed in the repo settings ... which I can confirm on my clone of the repo at https://github.com/uqs/OpenXcom/settings/actions  (default is indeed 90d)

But I have to back-pedal on this being useful for nightlies, as, well 90d isn't enough for that and you always get a zip-file, `chmod +x` gets lost :/. So good enough for Win builds, probably ok for MacOS as you'll have to do a bundle thing (and potentially sign it with a key??), but kinda meh for the Linux builds (but then again, Linux always comes with a compiler, which might be quite a tall order for an enthusiast on Windows systems to bootstrap). 
One could go with the `release` action, though, see https://github.com/actions/create-release

What I also dislike about them (for nightlies) is that while I can name them for the git hash, there's no way to get the shortened git hash. And using the long hash seems just silly, using `git describe` would need a round trip via a shell ENV. Hmm, let me play with that.

SupSuper:
I wouldn't worry about nightlies for now, just cleaning up the CIs. They suffer from random flukes so much I wouldn't really wanna depend on them anyways: https://github.com/OpenXcom/OpenXcom/runs/1356222879

Our nightly build server could use some cleaning as well but that's a whole different topic.

uqs:
402 Payment required, lolwut?

I'm happy with the git describe thing now, it produces an artifact with a meaningful name. So you can consider this branch ready to be merged. I've rebased it to the latest master, so you should be able to -ff merge it.

Once done, I'll rebase the travis CI branch and work on the suggestions there.

PR #1280, opened 2020-11-04
Build system cleanup

uqs:
Multiple build systems are confusing to people not familiar with the project and CMake is ported everywhere and is a Makefile-compiler, so it can spit out the proper Makefiles, ninja files or MSVC projects no problem.

Some of the non-CMake builds would also frequently break, as they are not checked in the CI, and can thus suffer bitrot.


SupSuper:
Overall looks good to me. Only things I would change are:
- The `Makefile.simple` requires zero maintenance so I would keep it, it's a simple solution when you just wanna compile the code without installing anything.
- The `README.OSX` shouldn't be merged with the main readme as it's primarily intended for developers. If anything it should be moved to the wiki: https://www.ufopaedia.org/index.php/Compiling_with_Xcode_(OpenXcom)

scarabeusiv:
We are using the autotools build system:
 https://build.opensuse.org/package/show/games/openxcom
Thats why i wrote it in first place. Also I can write you travis pr that will run it at least once in CI if you are afraid of regressions, I rebase it every couple of months so we are tracking the git head quite closely and would notice if something broke.

OTOH if it is too much to keep that one I will switch and check if it produces the same stuff.

scarabeusiv:
Oky I've rebased to cmake system https://build.opensuse.org/package/rdiff/games/openxcom?linkrev=base&rev=23
Seems to produce similar result with few minor changes we can live without.

uqs:
Umm, I'm confused, there's this file in the repo: install/opensuse/openxcom.spec and it contains:

```
...
BuildRequires:  cmake
...
%build
cd ..
mkdir temp
cd temp
cmake  -DBUILD_PACKAGE:BOOL=ON \
       -DDEV_BUILD:BOOL=OFF \
       -DCMAKE_INSTALL_PREFIX:PATH=/usr \
       -DCMAKE_BUILD_TYPE:STRING="Release" ../%{name}-%{version}/

make VERBOSE=1 %{?_smp_mflags}

```

So if OpenSUSE still used configure/autotools, that seems like a bug? I mean, this spec was added with e6d774182866c225216cfc3c51e9803e04ce5d5d back in 2012.

uqs:
@scarabeusiv  can Suse folks decide on a canonical spec file? The one inside openxcom itself or maybe your own copy of it (though debian, gentoo are also inside openxcom itself, so that seems more useful?)

scarabeusiv:
the spec you have in the folder here is definitely not used by openSUSE. But I've switched the package in the revision I linked above. Due to that my comment is now moot.

If you want you can pick up the .spec file from the above url as it should really build the package if somebody wishes to do so using rpmbuild.

uqs:
@SupSuper Makefile.simple restored; README merging removed again and also added a Windows build for Travis.

Please take another look, thanks!

scarabeusiv:
@uqs we never used the spec file that is present here. Debian and Gentoo are using the one from their respective trees too.

Some upstreams want to carry the distro build ruleset but they are not mostly used. I can assure you that the people using openSUSE will get the package from build.opensuse.org never touching the spec you have here in git anyway.

SupSuper:
@scarabeusiv brings up a good point, all the distros have their own build setup. I had a quick look around and found these:
[Fedora](https://copr.fedorainfracloud.org/coprs/evgenyz/openxcom/) - Autoconf
[Archlinux](https://aur.archlinux.org/packages/openxcom/) - Autoconf
[Gentoo](https://packages.gentoo.org/packages/games-engines/openxcom) - CMake
[Ubuntu](https://knapsu.eu/openxcom/) - CMake

How do we get everyone to play along? 😃

R1dO:
@scarabeusiv Just to make sure I understand your last comment correctly. 
Are you saying there is no point in keeping the stuff under `install` (at least for opensuse) since distro's will roll their own anyway?

scarabeusiv:
@R1dO honestly there is no need to keep it for anything there. Maybe if you have some special build to provide it, or if the distributions themselves would not ship it. But neither is the case now.

For the build the distributions will switch to cmake if the autotools disappear. The only item that is a bummer compared to the autotools is a missing manpage.

scarabeusiv:
Also it would be quite sensible when the buildsystem is decided to do a release. Not because of features, but because by that all the distributions would pick it up and ship using the new system.

uqs:
Fedora/Archlinux will indeed not break with the removal, they are still stuck on 1.0 anyway. And we can assume that package maintainers are very familiar with both autotools and cmake, as they are so ubiquitous. 

Fun fact, FreeBSD uses `Makefile.simple` :), see [this](https://github.com/freebsd/freebsd-ports/blob/master/games/openxcom/Makefile) :see_no_evil: 

@scarabeusiv could you clarify what the manpage has to do with autotools? The `openxcom.6` is using the very outdated roff macros (I can never recall which one is the old one). It really doesn't have much content, though an explanation for where to stick the datafiles surely is nice to have. I can write a proper new one, if it is deemed useful.

scarabeusiv:
Some distros like Fedora and Debian require all binaries to have manual page. Thus if they lose it they might be grumpy, the autotools automatically installed it in proper location, thats about it

I for one never felt the need to enforce this on openSUSE ;-)

SupSuper:
https://stackoverflow.com/questions/3375341/man-page-generation-packaging-installation-with-cmake

This might help?

scarabeusiv:
Yup that seems correct if it really works.

uqs:
I'm happy to take a look at the manpage handling in cmake once the other changes are in. There's also no reason this should require doxygen, manpages are rather simple to write.

I'd like to get this in first though. It won't break people right away and you can assign me a release-blocking issue for the manpage before the next release (maybe a bump to v1.1 is in order? or 2.0 if there's too much breakage in terms of data locations having moved around, I think).

uqs:
Thanks much, I've updated the wiki a bit, namely
https://www.ufopaedia.org/index.php/Compiling_(OpenXcom)
and https://www.ufopaedia.org/index.php/Compiling_with_CMake_(OpenXcom)

Someone would have to delete https://www.ufopaedia.org/index.php/Compiling_with_Autotools_(OpenXcom) as I have no permissions to do that.

uqs:
https://github.com/uqs/OpenXcom/commits/man for the manpage, I'll open a new pull request. It'd like to know if this is useful for package maintainers (it might be easier actually, if cmake does *not* try to install the manpage badly, i.e. uncompressed)

PR #1283, opened 2020-11-10
doc updates and man page rewrite

uqs:
Whether writing XML or not is probably a topic we could debate endlessly. Folks that care about manpages are probably fine with writing mdoc macros (I find them quite ok, once you get used to the mnemonics).

This allows downstream maintainers to drop the xmlto dependency

The fly in the soup is the shoddy manpage support by CMake, because it would have to compress the installed man page.

It would be good to hear whether the "forced" installation is useful, or whether package maintainers would rather copy&compress the manpage themselves in a post-install step.


uqs:
man page will render like so: https://i.imgur.com/BPDkiw6.png

R1dO:
Looks neat (rendered output).

Wondering though. Perhaps it is a good thing to mention that one need to own the original games (and copy datafiles to a required location) in order for the application to work. Possibly under description, or a separate caveat section.

Any thoughts on that?

P.s. The ``cmake -B build`` option will create that folder if it doesn't exist, no need to use `mkdir -p build` in the readme.

uqs:
Thanks for the build dir reminder, I must've mixed up something in my testing, most likely when I was dealing with configure. I've removed the extra mkdir's now.

I also added a short INSTALLATION section to the manpage, but I'm reluctant to copy the instructions from README over, as that just means we have it in 2 places and then stuff will diverge :/

R1dO:
I mix things up al the time :wink:

I agree with your point that copying instructions from the readme is less than ideal. From my point of view your current `Installation` section suffices. If you feel more is needed try the following suggestions (at your discretion)
- Add a link to https://www.ufopaedia.org/index.php/Installing_(OpenXcom)
- Consider renaming that section to WARNING. It stands out more (people might assume installation is already handled by the package manager).

As for your commit message on compressing manpages.
If you really want to gzip those, perhaps https://stackoverflow.com/questions/7050997/zip-files-using-cmake can be of help .
The only thing that we have to be careful about is not using an approach that requires a too recent version of cmake (there were some problems in the past because of that)

uqs:
gentle ping?

I think the manpage compression should be left to the package maintainers, there are usually tools to do that automatically. And WARNING as a section title seems rather drastic. There's already a link to the Wiki page as well.

SupSuper:
All the documentation needs an overhaul so I wouldn't worry about it for now.

docs/CMakeLists.txt
```diff
@@ -35,3 +35,5 @@ if ( NOT ( DOXYGEN_EXECUTABLE MATCHES "NOTFOUND" ) )
 else ()
   message ( "No doxygen command found. Disable API documentation generation" )
 endif ()
+
+INSTALL(FILES ${CMAKE_CURRENT_SOURCE_DIR}/openxcom.6 DESTINATION ${CMAKE_INSTALL_PREFIX}/man/man6)
```
MeridianOXC, line 39: This breaks my `make package` command after the build on OSX.
See result: https://imgur.com/a/UHfS10D
Build procedure: https://openxcom.org/forum/index.php/topic,7048.msg111477.html#msg111477

R1dO: Hmm yes ... that line is rather unconditional. Wouldn't be surprised if it breaks ``make package`` on windows as well.

I might be able to take a look at it later today to see if/how it can be solved.

Just one question though:
Do you want to keep the man-pages for the osx build [1] or can you live without that documentation option?
Reason: Due to lack of Apple coding knowledge it is probably easier to just disable that part.

[1] On first look apple seems to be aware of that concept

uqs: Sorry folks, I only know the FreeBSD packaging system, where I think this would all get "installed" into a staging area (so no permission problems there). 

Can you please pastebin the output of `make package VERBOSE=1` ?

uqs: This is "broken", if you will, under Linux as well. There, `make package` requires a `make install` first, which requires root permissions to put any of the files under /usr/.... So for Linux, you need to do `sudo make package` essentially. I've just tested this though, and only the binary ends up in the package, not the man page :(, it also misses bin/common and all the Readmes, so for non-macOS it seems pretty broken.

I have no idea how macOS works here, I guess you got some sort of package bundle? Do those ever contain man-pages?

I'd say to just remove that line again and package maintainers that want a man-page installed can do the needful.

@MeridianOXC could you also please test the macOS artifact, say here: https://github.com/OpenXcom/OpenXcom/actions/runs/383637098 ?
I've verified that both linux CI artifacts and the win build are producing working executables.

uqs: sorry for the spam, I got a package build going on the CI, but it produces a tarball, not a dmg file. Advice needed, thanks!

https://github.com/uqs/OpenXcom/runs/1495209792?check_suite_focus=true

R1dO: You probably need to incorporate many of the OXCE changes to CMakeList.txt for it to work. They put in a lot of effort to support modern osx and `.dmg` files. 
See: https://github.com/MeridianOXC/OpenXcom/blob/oxce-plus/CMakeLists.txt, especially line 184 which defines the generator to use when for packaging into `.dmg`.

For testing purposes you might get away with adding
`set ( CPACK_GENERATOR "DragNDrop")` to `cmake/modules/apple.cmake` 
But this is hypothetical (e.g. did not test it)

Looking at the commit you referenced in your last post: 
Am I correct in observing that your solution to the problem encountered by Meridian is to remove the offending line? Under the rationale that the only users that would benefit from it *might* be "distribution package maintainers", hence it is not worth the problems for normal releases.

MeridianOXC: I've removed the line for OXCE, just wanted to report back here that others may have the same issue.

uqs: Yes, I think we should either remove or at least skip it for Apple/Windows, I'm testing a change for that now. See my initial message on the pull request where I wasn't sure that this would be useful :(

`make install` also uncovered an existing bug with the MSVC setup, I guess the windows builds are never actually done with cmake, so no wonder this got broken.

uqs: https://github.com/OpenXcom/OpenXcom/pull/1290

(stupid travis' macos has homebrew problems again ...)

PR #1285, opened 2020-11-11
Fix for RelWithDebInfo behaving as debug build

### LINKED ISSUE #1284: cmake .. -DCMAKE_BUILD_TYPE=RelWithDebInfo behaves like full debug build (12 FPS)
After latest changes in cmake compilation the cmake configuration option:
cmake .. -DCMAKE_BUILD_TYPE=RelWithDebInfo behaves like a full Debug build. Meaning this is unplayable with 12 FPS or so.
Please make sure to revisit changed procedures and make -DCMAKE_BUILD_TYPE=RelWithDebInfo
Playable again :)

R1dO:
The empty "add_compile_options()" seem to clear all other compile flags. Removing them should not hurt.
Should fix #1284 

Image shows CMakeCache.txt diff. Left is before this commit, right is after (and equal to behavior before pr #1280 )
Screenshot_2020-11-11_18-51-25

--edit--
Fix PR description (title and text link to issue)

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -374,8 +374,6 @@ if ( "${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" OR CMAKE_CXX_COMPILER_ID MATCHES 
       )
     if ( "${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" )
       # add warning flags recognized by g++ but not by clang
-      add_compile_options (
-	  )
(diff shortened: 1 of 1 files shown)


R1dO:
Retracting this PR.

It does not seem to have anything to do with the linked issue (which has solved itself).

Even worse, after further investigation the error as shown in the image was caused by an incomplete build-system. The only reason the error was fixed on my end is due to fixing the build-system not the PR.

PR #1290, opened 2020-12-05
Install the manpage only for Linux/*BSD, run `make install` during CI runs

uqs:
Reported by: @MeridianOXC

The manpage is pretty useless under Windows and also for the macOS `dmg`
archives. The `make install` target could also use some more work, but
at least it's working now.

CPack could potentially be used, but would need a lot of work and the
current GH artifact build step is actually superior.

While here, use the proper NPROC value for all runners (except windows,
which falls back to 2). Linux needs sudo to install, Win/Mac don't.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,7 +73,10 @@ jobs:
         shell: bash
         run: |
           echo "describe=$(git describe)" >> $GITHUB_ENV
-
+          if [ "${{ matrix.name }}" = "Linux" ]; then
+            echo "MAYBE_SUDO=sudo" >> $GITHUB_ENV
+          fi
+          echo "CMAKE_BUILD_PARALLEL_LEVEL=`getconf _NPROCESSORS_ONLN 2>/dev/null || getconf NPROCESSORS_ONLN 2>/dev/null || echo 2`" >> $GITHUB_ENV
       - name: Download dependencies (Windows)
         if: runner.os == 'Windows'
         run: |
@@ -83,7 +86,6 @@ jobs:
 
       - name: Generate project files
         run: |
-          mkdir ${{ matrix.build-dir || 'build' }}
           cmake -B ${{ matrix.build-dir || 'build' }} ${{ matrix.build-src-dir || '.' }} -DCMAKE_BUILD_TYPE=Release -DENABLE_WARNING=1 -DCHECK_CCACHE=1 ${{ matrix.cmake-args }}
         env:
           CC: ${{ matrix.compiler }}
@@ -91,7 +93,16 @@ jobs:
 
       - name: Compile source code
         run: |
-          cmake --build ${{ matrix.build-dir || 'build' }} -v -j2 --config ${{ matrix.build-config || 'Release' }} -- ${{ matrix.build-args }}
+          cmake --build ${{ matrix.build-dir || 'build' }} -v --config ${{ matrix.build-config || 'Release' }} -- ${{ matrix.build-args }}
+
+        # Note, this is bogus on MacOS, as it installs shaders et al twice.
+        # Windows installs under C:/Program Files (x86)/OpenXcom/bin (where the
+        # 'bin' is pretty redundant) and Linux installs under various dirs under
+        # /usr/local
+      - name: Make install
+        shell: bash
+        run: |
+          $MAYBE_SUDO cmake --build ${{ matrix.build-dir || 'build' }} -v --config ${{ matrix.build-config || 'Release' }} --target install
 
       - name: Archive artifacts
         uses: actions/upload-artifact@v2
--- a/docs/CMakeLists.txt
+++ b/docs/CMakeLists.txt
@@ -36,4 +36,7 @@ else ()
   message ( "No doxygen command found. Disable API documentation generation" )
 endif ()
 
-INSTALL(FILES ${CMAKE_CURRENT_SOURCE_DIR}/openxcom.6 DESTINATION ${CMAKE_INSTALL_PREFIX}/man/man6)
+# Only useful for make install/package under *nix OSes, but not macOS.
+if ( UNIX AND NOT APPLE )
+    INSTALL(FILES ${CMAKE_CURRENT_SOURCE_DIR}/openxcom.6 DESTINATION ${CMAKE_INSTALL_PREFIX}/man/man6)
+endif ()
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -481,10 +481,6 @@ if ( WIN32 )
     set ( static_flags  -static )
     set ( SDLMIXER_LIBRARY "${SDLMIXER_LIBRARY} -lwinmm" )
   endif ()
-  if ( MSVC )
-    set_target_properties ( openxcom
-      PROPERTIES PREFIX ../ )
-  endif ()
   set ( system_libs ${basic_windows_libs} SDLmain ${static_flags} )
 endif ()
 


uqs:
Travis's MacOS image is stuck 52min in homebrew bootstrapping ... again :(, ignore that.

.github/workflows/ci.yml
```diff
@@ -91,7 +94,16 @@ jobs:
...
       - name: Compile source code
         run: |
-          cmake --build ${{ matrix.build-dir || 'build' }} -v -j2 --config ${{ matrix.build-config || 'Release' }} -- ${{ matrix.build-args }}
+          cmake --build ${{ matrix.build-dir || 'build' }} -v --config ${{ matrix.build-config || 'Release' }} -- ${{ matrix.build-args }}
+
+        # Note, this is bogus on MacOS, as it installs shaders et al twice.
+        # Windows installs under C:/Program Files (x86)/OpenXcom/bin (where the
+        # 'bin' is pretty redundant) and Linux installs under various dirs under
+        # /usr/local
+      - name: Make install
```
R1dO, line 103: Is there any benefit for installing the game inside the CI?

As far as i know the CI exist to ensure code changes do not break compilation on other OS's, it is not used to create packages (or play via github's CI servers :wink: ).

.github/workflows/ci.yml
```diff
@@ -91,7 +94,16 @@ jobs:
 
       - name: Compile source code
         run: |
-          cmake --build ${{ matrix.build-dir || 'build' }} -v -j2 --config ${{ matrix.build-config || 'Release' }} -- ${{ matrix.build-args }}
+          cmake --build ${{ matrix.build-dir || 'build' }} -v --config ${{ matrix.build-config || 'Release' }} -- ${{ matrix.build-args }}
```
R1dO, line 97: You might want to put `-j`  just before `-v`. That instructs cmake to tell the build tool we want parallel compilation. In the case of `make` that implies "being greedy" (e.g. use  as much jobs as possible). In that case you would not need line 79.

.github/workflows/ci.yml
```diff
@@ -73,7 +73,10 @@ jobs:
         shell: bash
         run: |
           echo "describe=$(git describe)" >> $GITHUB_ENV
-
+          if [ "${{ matrix.name }}" = "Linux" ]; then
+            echo "MAYBE_SUDO=sudo" >> $GITHUB_ENV
+          fi
+          echo "CMAKE_BUILD_PARALLEL_LEVEL=`getconf _NPROCESSORS_ONLN 2>/dev/null || getconf NPROCESSORS_ONLN 2>/dev/null || echo 2`" >> $GITHUB_ENV
```
R1dO, line 79: If using the `-j` flag (see commit on line 97) this one is not needed anymore. Less variables to worry about :wink: .

And while i am still in a clean-up mood: Consider deleting line 89 (`mkdir ...`) because `cmake --build ...` should take care of that.

.github/workflows/ci.yml
uqs: Yes, the benefit is testing whether changes to CMakefiles break people's workflow, like the manpage addition did. So let's not do that again :)

Also, the artifacts that this build produce work fine, so one could point the nightly builds over to the github servers, just saying :)

Furthermore, ideally we don't just build things, but actually run some end to end tests ... but of course we can't ship the gamefiles anywhere, so these tests would be rather theoretical (like maybe the one savegame crash report that came in earlier today)

.github/workflows/ci.yml
uqs: Ah, that seems to be a GNU extension, it didn't always do that and a number after -j was required.

But then I would have to test whether the MS build system and the Apple one also do the right thing. Would be interesting to see what happens in a CI run for those ... *wink* *wink* *nudge* *nudge*

.github/workflows/ci.yml
uqs: I'm looking forward to seeing your pull request for this and the CI runs' output on non-Linux systems ;]

mkdir removed, will upload new snapshot soon, thanks for pointing that out.

.github/workflows/ci.yml
R1dO: That is a good enough explanation for me ;)

.github/workflows/ci.yml
R1dO: I mentioned the `-j` flag because that offloads the decision making logic to cmake so that we do not have to perform OS specific checks (kinda the reason we use that build system in the first place).

Off course i can always do a PR, but not at the moment. You kinda took on this project and by PR'ing right now things can become messy pretty quick. I don't want to get in your way atm.

Another tip:
If using the `-DCMAKE_INSTALL_PREFIX=~/.local` or `-DCMAKE_INSTALL_PREFIX=$HOME/.local`  during "Generate project files", you might even get do without requiring `sudo`.

uqs: Consider me sufficiently nerd sniped, but let's get this fix in first :) 

Slight correction though, `cmake` doesn't do any magic here, it's all the result of `GNU make` handling `-j` sufficiently well, which I didn't know before!

During a run:
```
% ps -efH|grep -A3 "[c]make"
uqs       427678  247986  3 15:26 pts/10   00:00:00                 cmake --build build-linux -v -j
uqs       427679  427678  0 15:26 pts/10   00:00:00                   /usr/bin/cmake -E env VERBOSE=1 /bin/gmake -j
uqs       427680  427679  0 15:26 pts/10   00:00:00                     /bin/gmake -j
uqs       427683  427680  0 15:26 pts/10   00:00:00                       /bin/gmake -f CMakeFiles/Makefile2 all
```

See how cmake does not figure out the wanted parallelism at all, it just hands that over to gmake. And well, that needs testing on Mac/Windows, as those don't use GNU make :) (and I use neither of those systems)

uqs:
Uhh, don't try this at home kids. I had my laptop grind to a halt, trying a build with `-j` (sans number). It spawned as many clang's as it could, causing excessive swapping, naturally.

From GNU make's man page:
```
       -j [jobs], --jobs[=jobs]
            Specifies the number of jobs (commands) to run simultaneously.  If there is more than one -j option,
            the last one is effective.  If the  -j  option  is  given  without  an argument, make will not limit the
            number of jobs that can run simultaneously. When make invokes a sub-make, all instances of make
            will coordinate to run the specified number of jobs at a time; see the section PARALLEL MAKE
            AND THE JOBSERVER for details.
```

So uuh, this can be super useful in some projects, but not when all you do is forking C++ compilers that gobble up all the RAM they can get their hands on.

Of course I tried this anyway (not here though, a different branch), and the funny thing is:
- this works fine under macOS, compile time is 9m
- this works fine under Windows, compile time is 8m
- this FAILS for both Linux builds after 19--23m, the workers run out of RAM

https://github.com/uqs/OpenXcom/actions/runs/402658186

:)

.github/workflows/ci.yml
uqs: oops, I left a comment in the other thread. tl;dr don't use `-j` on github runners.

R1dO:
Ouch that is indeed rather nasty.

Weird thing though, i never encountered this error (or swapping). Both on my own system [1] and the test runs i did on github: https://github.com/R1dO/OpenXcom/actions/runs/402759753
But since it happened to your build (and i saw no obvious setup errors there) the approach of limiting is probably the smart thing to do (safer in case there are some runner hickups).

Some extra things to notice from my tests:
* Windows compile does not seem to pickup `-j`, it requires a C or CXX flag (and a dev with windows experience :wink:).
  - Simply putting that flag the way i did (without specifying a number) actually leads to worse compile time. 
    It limits to one core only, `/MP1` in screenshot:
    image
* You can do without `sudo` if specifying the install prefix in the github action.
  - My approach was to put `cmake-args: -DCMAKE_INSTALL_PREFIX=$HOME/.local` in the build matrix for linux and Apple.  
    Perhaps you have a better solution (just make sure it does not enter the windows build, that will fail).
* Could not test on travis though, somehow it does no longer recognizes my fork.


[1] Entirely possible because it is too beefy for my own good, which works well with greedy applications.

SupSuper:
CMake doesn't convert between compiler flags, so any flag that works on GNU/Clang will obviously not work on MSVC, and vice-versa.

`-j` should never be used without a number and I don't understand why that's even an option, it will always do something dumb like suck up all your resources, crash or "successfully" compile a broken executable. The convention is `-j<number of cores>`.

`/MP` is a project compiler flag and should just be added directly to CMakeLists if you wanna use it: https://stackoverflow.com/questions/33858455/how-to-configure-cmake-in-a-way-that-it-generates-visual-studio-2013-project-fi
I don't know why it defaults to MP1 when passed like that, probably because it only sees one source file.

R1dO:
Good discussion here.  Time for me to get rid of this bad behavior (-j).

Based on uqs's test and remarks from both i did some more background reading. Guess i was to naive about (c)make smartness when determining core/memory limits :shrimp: .

PR #1294, opened 2020-12-17
snap new base to the nearest city in XCOM1

feiyunw:
Paranoids like me always want to have bases built at the very same point of a city in XCOM1.
Unfortunately, the base naming dialog hasn't a cancel button, neither will it respond to ESC button, thus make it somehow cumbersome.
The mobile version makes it more difficult.
I added a new feature to automatically snap new base to the nearest city in XCOM1 when enabled.

### DIFF
--- a/bin/standard/New_Base_In_City/New_Base_In_City.rul
+++ b/bin/standard/New_Base_In_City/New_Base_In_City.rul
@@ -1,0 +1,3 @@
+# New_Base_In_City
+constants:
+  - newBaseInCity: true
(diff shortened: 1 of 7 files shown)


Player701:
Hmm... Sorry if I don't understand, but why was this closed?

SupSuper:
Nobody showed any interest.

Player701:
Ah, I see. Well, I sometimes like building my bases in cities too, although I haven't played in a while, and I don't always have the time to comment on everything. But I think this could've worked better as an UI extension (e.g. hotkey or button toggle) instead of as a mod.

PR #1303, opened 2021-01-19
fix statistics calculation for commendations

just-dont:
Addresses https://github.com/OpenXcom/OpenXcom/issues/1302 and https://github.com/OpenXcom/OpenXcom/issues/1287

* woundsHealedTotal wasn't tallying totals
* killsWithCriteria<...> commendations were tallying results awkwardly, getting tallies for each OR criteria block separately from other blocks, which resulted in "best of" tallying across all OR blocks, rather than "total of" which was the presumed behavior by Commendations mod logic as well as other mods such as X-Piratez

This code produces mostly the same end result as what @karadoc did [here.](https://github.com/karadoc/OpenXcom/commit/a6055080d2f38602191fe3d5a3c4dd8329cd5854) The difference is in algorithms, as well as addressing https://github.com/OpenXcom/OpenXcom/issues/1287, as they both are about commendation statistics bugs.


SupSuper:
@Shoes01 Any opinion on this?

Shoes01:
Sorry I'm late to the party :)

You're right that the code, as written, was not mean to total up the "Or Criteria". That was intended, but in retrospect that doesn't make much sense. I think this change is a net improvement.

Issue #1304, opened 2021-01-27
Playable Aliens mod incompatible with OpenXcom


Tamamo-no-Bae:
I wasn't sure where to post this, but the original Playable Aliens mod I made a long long time ago for the originals seem to be incompatible with OpenXcom.
They cause a crash, I will post the mod packs and a crash dump, I hope this can be fixed in a future version.

In the original X-Com, Aliens are defined as armor. Thus having an armor value above the default changes you into an Alien.
Likewise Aliens use an unused internal rank called "Units in Entrance"

This can be seen in the TFTD save with the accurate Tasoth Soldier recreation or if you used the Tentaculat/Chryssalid bug to get one on your team.

So why does this require a mod? Because inventory data for those armor are missing, thus causing a crash.
However in Openxcom it's possible the support for the Alien armors are missing entirely. 

Here's hoping this can be fixed!
[openxcom.log](https://github.com/OpenXcom/OpenXcom/files/5877544/openxcom.log)

[pack3.zip](https://github.com/OpenXcom/OpenXcom/files/5877547/pack3.zip)
[pack.zip](https://github.com/OpenXcom/OpenXcom/files/5877548/pack.zip)

MeridianOXC:
This is not a bug and also not a missing functionality.
Everything works as expected and as intended.

We may consider giving the user a nicer error message with some explanation, instead of crashing... but it won't be possible to load these modded saves. I will look into that.

The correct way to introduce this such modded behavior is to use OpenXcom's own modding capabilities.

MeridianOXC:
Assigning to SupSuper as he wanted to implement some kind of support for this.

PR #1306, opened 2021-02-07
Graceful exit on modded OG saves

MeridianOXC:


SupSuper:
Isn't this the opposite of what #1304 wanted? :P

MeridianOXC:
> 
> 
> Isn't this the opposite of what #1304 wanted? :P

Yes, it is the opposite.
I have explained why the bugreport is invalid and cannot be implemented.
I have also given him a guideline (on the forum) how to implement it properly via OpenXcom modding.

This commit is just about giving the player a "nicer" crash with explanation of why it is crashing.

PR #1307, opened 2021-02-07
More precise distance calculation for 2x2 units

MeridianOXC:
Updated ONLY the calculations related to maxRange attribute of ranged weapons.
These calculations don't exist in the OG, relevant only for mods.

All other distance calculations were unchanged,
difference of +/-1 doesn't mean much in them anyway.


SupSuper:
On the surface it looks fine but @Warboy1982 should probably double-check this.

PR #1317, opened 2021-03-08
Fix wasted fuel bug

MeridianOXC:


SupSuper:
This doesn't address the core issue, which is the "refuelling" phase always consumes an item before checking if it's necessary. So the bug might still happen if refuelling is triggered some other way.
I've fixed it in https://github.com/OpenXcom/OpenXcom/commit/adb97235b1566440223321fb4b7eaa0a47db7ed7 ands also moved the whole logic to Craft, which will probably need patching up on the OXCE end but hopefully makes it cleaner.

PR #1322, opened 2021-03-15
First step: AI vs. AI

fimmtiu:
* Embeds system-installed Janet library.
* Adds a stub Janet AI module which at present just runs the normal AI code.
* Fixes some hard-coding in the AI which won't let two AI factions fight.
* Sets all X-Com soldiers to use the JanetAIModule.
* Sets the `traceAI`, `skipNextTurnScreen`, and `noAlienPanicMessages` options by default so that (a) you get a lot of useful AI output, and (b) you don't have to interact with the battle as much by clicking to dismiss things.

Now you can start a battle and both sides will fight each other.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ Makefile.in
 depcomp
 compile
 .deps/
+.vscode
 *.tar.gz
 *.tar.xz
 openxcom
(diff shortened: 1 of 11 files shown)


fimmtiu:
Crap, sorry! This was supposed to be pushed to my private fork. Not an actual pull request for the main OpenXcom project.

PR #1337, opened 2021-06-16
Typo fix (fixes unability to launch TFTD)

hamer:
A minor fix (possibly a mistyping) that caused TFTD fail to start

### DIFF
--- a/bin/standard/xcom2/vars.rul
+++ b/bin/standard/xcom2/vars.rul
@@ -18,7 +18,7 @@ alienFuel: [STR_ZRBITE, 50]
 fontName: Font.dat
 
 constants: #done
-  doorSound: 47
+- doorSound: 47
   slidingDoorSound: 46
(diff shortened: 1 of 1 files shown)


SupSuper:
Exactly what bug does this fix? `constants` is a map, not a vector, this will just make it ignore all the variables.

hamer:
It is attaempt to fix the error message:

```
Failed to load 'X-Com: Terror From the Deep'; mod disabled for next startup
/usr/local/share//openxcom/standard/xcom2/vars.rul: invalid node: this may result from using a map iterator as sequence iterator, or vice-versa
```

that's why that assumption was done.
anyway, sorry for disturbing. this evening i'll dive deeper in code to understand what happens

hamer:
The bug does not exist and therefore this pullrequest is revoked.
The reason of this behavior was older binary located in $PATH

Issue #1266, opened 2020-09-05
lowAccuracyHitCounter never increases


Delian0:
There is a soldier statistic called lowAccuracyHitCounter that is used for commendations. It doesn't increase correctly if the range-reduced accuracy option is turned on - the statistic uses standard accuracy calculation instead.

In other words, even if I have 0% accuracy and I hit an enemy, this statistic won't increase.

Shoes01:
The relevant code for increasing the counter is:
```c++
if (_unit->getFiringAccuracy(_action.type, _action.weapon) < _parent->getTileEngine()->distance(_action.actor->getPosition(), victim->getPosition()))
{
	_unit->getStatistics()->lowAccuracyHitCounter++;
}
```
where `_unit->getFiringAccuracy(_action.type, _action.weapon)` is the accuracy of the weapon, and `_parent->getTileEngine()->distance(_action.actor->getPosition(), victim->getPosition())` is the distance between the actor and the victim. 

The firing accuracy is the result from [here](https://github.com/OpenXcom/OpenXcom/blob/7aae3f45365d4636b40e05ac80f6ebd2575efb59/src/Savegame/BattleUnit.cpp#L1506), which as far as I can tell should be accurate.

I can think of only the following as a potential bug: the commendation is broken in some other way, or the firing accuracy shown in the menu is different than what is actually used.

PR #1344, opened 2021-09-25
lowAccuracyHitCounter fix

Delian0:
lowAccuracyHitCounter should take battleUFOExtenderAccuracy into account

Closes issue #1266

I have not tested this code.


MeridianOXC:
the bugreport is valid 
the fix is correct... but ugly
at minimum, it needs a squash
(at best, I'd like to see the calculation extracted into a separate method... but... small steps)

PR #1360, opened 2022-03-12
FastVoxelTraversal algorithm attempt 2, success !

SkybuckFlying:
This is not really a pull request, but a way to bring it to your attention.

The bresenham line algorithm has been successfully replaced by this fast voxel traversal algorithm. It does not have yet any clipping of input line and it should to make it faster. It also checks for negative z and skips these.

It's interesting code though, it reduces the number of voxel processing sites to just 2 instead of 3.

Perhaps once clipping code is added it may also run more efficiently then the bresenham 3d calculateLine method ?  Not sure, will examine the old code later on to see how much difference it makes...

I CAN CONFIRM THE FAST VOXEL TRAVERSAL ALGORITHM IS A LOT MORE ACCURATE THEN THE BRESENHAM 3D LINE ALGORITHM, AT LEAST WHEN doVoxelCheck := false;

EVEN WITH doVoxelCheck := True; THE BRESENHAM 3D LINE ALGORITHM WILL STILL MISS SOME CELLS !

VERY INTERESTING DISCOVERY !

(If you do pull it in I hope you make some further improvements :) like maybe add line clipping to grid or something and maybe check that it fully works within the boundarys of the map, I am not sure what happens if Z becomes very big, maybe it might still crash the game.)

I AM CONVINCED THIS INCREASED ACCURACY HAS A HUGE EFFECT ON THE GAME, QUITE FRANKLY I IMMEDIATELY NOTICED IT.

BETTER REMOVAL OF FOG OF WAR/BETTER VISUALS/SPOTTING.

MORE ACTIVE AI FOR ALIENS AND HUMANS, THEY CAN ACTUALLY SEE US BETTER NOW.

BETTER FIRING BETWEEN WALLS, QUITE AMAZING AND EXCITING, I AM THINKING OF RE-PLAYING THIS GAME WITH THIS "MOD" CAUSE IT JUST MADE THE GAME A WHOLE LOT MORE INTERESTING.

ALSO THE RELEASE WIN32 BUILD RUNS FINE ON MY SYSTEM NO NOTICEABLE SLOW DOWNS.

(NOT SURE ABOUT WIN64 BUILDS, NO LIBS FOR IT, WOULD HAVE TO RE-BUILD THOSE, I JUST USED A PRE-BUILD DEPENDANCY PACKAGE FOR THE DLLS AND SUCH, PERHAPS THIS GAME IS NOT READY YET FOR 64 BIT COMPILATION ;))

WARNING: CURRENT IMPLEMENTATION STILL CAN CAUSE HANGS, IT'S NOT THE ALGORITHM BUT SOMETHING ELSE, POSSIBLY NO CLIPPING/GOING OUT OF RANGE OF MAP, OR NO NEGATIVE Z VALUES CAUSE OF Z CAP, DEBUG IT WELL FIRST ! ;) (OR OPENXCOM HAS OTHER BUGS ?!?) OTHER POSSIBILITIES COULD INCLUDE DOORS/OPENING CLOSING, ALIENS ON DIFFERENT LEVELS OR PERHAPS PATH FINDING BUG.

THE INCREASED ACCURACY IS WORTH PURSUEING THIS FURTHER AND GETTING THIS IMPLEMENTATION TO WORK FLAWLESSLY. 

*Update*:

I believe calculateLine is now fixed by making sure voxel and map checks stay within voxel boundaries.

Max voxel boundary position calculated as follows:
MaxVoxelX = ((MapSizeX-1) * 16)-1
MaxVoxelY = ((MapSizeX-1) * 16)-1
MaxVoxelZ = ((MapSizeX-1) * 24)-1

*** Inserting Update ***
Problem found, this calculation is wrong and should be:

MaxVoxelX = (MapSizeX * 16)-1
MaxVoxelY = (MapSizeX * 16)-1
MaxVoxelZ = (MapSizeX * 24)-1

This is because coordinates that go into CalculateLine are "voxel positions" across the entire map, and are not tile based positions ;)
****************************

I believe a tile is 16x16x24, if this is incorrect than please fix and correct it ! :)

calculateParabole was popping trajectory while trajectory could be empty !

Check for !empty added.

Oh that reminds me, some floats in calculateLine was changed to double for maximum accuracy, though since maps so small this can probably be changed to floats for faster code execution.

But first I want to produce some good first person view PNGs to compare if it makes any difference =D

My hypothesis why the original bresenham line algorithm can't see so far is because of the drifting and it causes inaccuracies, it might collide early against a wall/tile which it should not have collided with it.

This could explain why fast voxel traversal algorithm can see further ! and hence this explains why the visibility/spotting may be better and also why the human and alien AI react better ! =D

Basically it changes the game ! Very FAT=SUPERUBERCOOL ! =D

See evidence below, screens have been produced by saving/loading game and switch to different version of xcom, one with bresenham and one with fast voxel traversal ! the evidence is clear though not yet proven without a reasonable doubt, but I am convinced, proving it 100% could be done by colorizing where the algorithms end, interesting !! =D:
(This idea has not yet been done, first I want to have some fun with this new version, beta test it some more and then maybe ! ;):))

fpslook666BresenhamAlgorithmExample1
fpslook666FastVoxelTraversalAlgorithmExample1
fpslook667BresenhamAlgorithmExample2
fpslook667FastVoxelTraversalAlgorithmExample2
fpslook668BresenhamAlgorithmExample3
fpslook668FastVoxelTraversalAlgorithmExample3
fpslook669BresenhamAlgorithmExample4
fpslook669FastVoxelTraversalAlgorithmExample4

SOMETIMES IT SAYS: "NO LINE OF FIRE" BUT WHEN I KEEP CLICKING EVENTUALLY IT WILL FIRE, NOT SURE YET WHAT IS CAUSING THIS, COULD BE AN ALREADY EXISTING BUG IN THE GAME, NOT SURE, OR MAYBE SOMEHOW THE GAME CANNOT SET THE COORDINATES ON CURSOR CORRECTLY ?!?

BETA TESTING REVEALED ANOTHER PROBLEM: ON BASE DEFENSE MAP THE UNITS ON THE UPPER FLOOR CAN SEE THROUGH THE FLOOR AND CAN SEE ALIENS IN THE HANGER, HOW IS THIS POSSIBLE ?! OH MY GOD ! NEED FIX ! SHIT THE ORIGINAL BRESENHAM CODE DOES NOT HAVE THIS PROBLEM, WHAT THE FUCK ?! WHY ?! BOOEHOEE ! TO BE DEBUGGED/CONTINUED ! =| ANY HELP BE GREAT ! SO I CAN MOVE ON TO OTHER THINGS LIKE SHADING ! =|

*** 25 MARCH 2022 *** UPDATE:

CODE IS GETTING BETTER AND BETTER, ALMOST READY FOR A REAL PULL REQUEST, STILL ONE MORE BUG FIX INCOMING FOR LAST VOXEL POSITION CALCULATION, NEW CODE ADDED WHICH CLIPS POSITION,TARGET BASICALLY LINE TO GRID, GOOD HEAVILY TESTED CODE, BUT STILL ONE LITTLE BUG FIX COMING.

CODE COULD ALSO USE SOME REFACTORING. 

I  CAN TO DO THINGS: EITHER LET THIS ONE BE COMMITED.

OR I CAN GREAT A NEW FEATURE BRANCH WITH SUPER CLEAN CODE AND SUPER CLEAN COMMITS, THE CHOICE IS YOURS. THOUGH IT MAY NOT NEVER BE FULLY CLEAN, CAUSE NEW BUGS OR CODE OPTIMIZATIONS MIGHT HAPPEN... HMMM STILL A BIT EARLY TO DECIDE IF IT'S TIME FOR A MERGE ;)

*** UPDATED 2 ON MARCH 25-3-2022 ***

vMapVoxelBoundaryMaxX computed correctly now, now commit added.

Code should be pretty good right now !

I do notice some strange bugs with time units, and guns shooting again, but to me this seems bugs elsewhere in the code base and not related to my code changes.

Also in debug mode after all my units are moved, new time units are handed out ?

### DIFF
--- a/src/Battlescape/TileEngine.cpp
+++ b/src/Battlescape/TileEngine.cpp
@@ -3275,6 +3275,34 @@ bool LineSegmentIntersectsBoxSingle
 	return result;
 }
 
+/*
+
+Debug mode
+
+Debug mode is a special feature that makes it easier for developers/contributors to test out game features without needing to play a new game from scratch. After setting "debug: true" in the Options File, you can enable it by pressing Ctrl-D in-game to access a variety of features:
+
+Geoscape: Unlocks all research and shows globe country and region borders. Clicking on the globe returns the point coordinates. This option can be toggled, and toggles between showing country borders, regional borders, and mission zones.
+
+Battlescape: Whole map is revealed and opening the inventory screen resets TUs. Clicking the map returns the point coordinates. With traceAI enabled in the config, AI behavior is revealed and logged. Please note, in debug mode QuickSave/QuickLoad feature in Battlescape are completely disabled.
+
+Ctrl-V will reset visibility on all tiles, useful for observing LOS behaviour.
+
+Ctrl-K will kill all live aliens in the mission (nothing will be recovered).
+
+Ctrl-J will stun all live aliens in the mission (useful for testing recovery).
+
+Ctrl-W will warp the selected unit to the mouse cursor's current position.
+
+F8 will cycle through various engine processing speeds, useful for debugging animation errors.
+
+You can also press Ctrl-U or set "debugUi: true" to reveal the borders of the interface text, to debug translation errors.
+
+If the Commendations Mod is being played, Ctrl-C will wipe all earned commendations while retaining statistics, useful for debugging how they're awarded. 
+
+
+*/
+
+
 int TileEngine::calculateLine(Position origin, Position target, bool storeTrajectory, std::vector<Position> *trajectory, BattleUnit *excludeUnit, bool doVoxelCheck, bool onlyVisible, BattleUnit *excludeAllBut)
 {
 	float tMaxX, tMaxY, tMaxZ, t, tDeltaX, tDeltaY, tDeltaZ;
@@ -3366,28 +3394,29 @@ int TileEngine::calculateLine(Position origin, Position target, bool storeTrajec
 		)
 		{
 			// copy code later
-			cx = x1;// / TileWidth;
-			cy = y1;// / TileHeight;
-			cz = z1;// / TileDepth;
+			cx = x1;
+			cy = y1;
+			cz = z1;
 
 			// process voxel
 //			ProcessVoxel
 //			(
-//				trunc(x1 / mCellWidth),
-//				trunc(y1 / mCellHeight),
-//				trunc(y1 / mCellDepth)
+//				trunc(x1),
+//				trunc(y1),
+//				trunc(y1)
 //			);
 
-			if (_save != 0)
-			{
-				fpos = Position( cx, cy, cz );
-				fTile = _save->getTile(fpos);
-
-				if (fTile != 0)
-				{
-					fTile->setSmoke( 250 );
-				}
-			}			
+			// Skybuck: debug code, uses green smoke, blinds everything on accident
+//			if (_save != 0)
+//			{
+//				fpos = Position( cx, cy, cz );
+//				fTile = _save->getTile(fpos);
+//
+//				if (fTile != 0)
+//				{
+//					fTile->setSmoke( 250 );
+//				}
+//			}			
 
 			// process voxel
 			{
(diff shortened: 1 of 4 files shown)


src/Battlescape/TileEngine.cpp
```diff
@@ -2493,10 +2494,10 @@ int TileEngine::calculateLine(Position origin, Position target, bool storeTrajec
 		(
 			(cx >= 0) &&
 			(cy >= 0) &&
-			(cz >= 0)// &&
-//			(cx < _save->getMapSizeX()) &&
-//			(cy < _save->getMapSizeY()) &&
-//			(cz < _save->getMapSizeZ())
+			(cz >= 0) &&
+			(cx < _save->getMapSizeX() * 12) &&
```
Warboy1982, line 2498: should be 16

src/Battlescape/TileEngine.cpp
```diff
@@ -2493,10 +2494,10 @@ int TileEngine::calculateLine(Position origin, Position target, bool storeTrajec
...
+			(cy < _save->getMapSizeY() * 16) &&
+			(cz < _save->getMapSizeZ() * 16)
```
Warboy1982, line 2500: should be 24

SkybuckFlying: That code is disabled/old code and can be removed.

SkybuckFlying:
My official pull request is here:

https://github.com/OpenXcom/OpenXcom/pull/1373

Closing this unofficial pull request.

PR #1380, opened 2022-06-01
More exact palette full range 0..255

SkybuckFlying:
Full discussion here:

https://openxcom.org/forum/index.php/topic,10588.0.html

Following formula is used to convert 6-bit vga palette to 8-bit rgba palette as accurately as possible:
Calculations done with doubles for maximum accuracy during palette loading.
"Used computer variable for "fun code""

NewR = round( (OldR * 255.0) / 63.0 );
NewG = round( (OldG * 255.0) / 63.0 ); 
NewB = round( (OldB * 255.0) / 63.0 );

Have fun with it ! =D

### DIFF
--- a/src/Engine/Palette.cpp
+++ b/src/Engine/Palette.cpp
@@ -19,6 +19,7 @@
 #include "Palette.h"
 #include <fstream>
 #include "Exception.h"
+#include <cmath>
 
 namespace OpenXcom
 {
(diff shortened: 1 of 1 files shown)


blinkdog:
Perhaps this could work?

```cpp
_colors[i].r = ((value[0] << 2) | (value[0] >> 4)) & 0xff;
_colors[i].g = ((value[1] << 2) | (value[1] >> 4)) & 0xff;
_colors[i].b = ((value[2] << 2) | (value[2] >> 4)) & 0xff;
```

PR #1392, opened 2022-10-15
Fix DATADIR compile option

### LINKED ISSUE #1391: -DDATADIR has no effect
#### What I would expect to happen:
Data files get installed to `${CMAKE_INSTALL_FULL_DATADIR}/openxcom`
The game searches the following locations for data files:
- `$XDG_DATA_HOME/openxcom` (if `$XDG_DATA_HOME` is set)
- `$HOME/.local/share/openxcom` (if `$XDG_DATA_HOME` isn't set)
- `${DATADIR}`
- `$XDG_DATA_DIRS/openxcom` (if `$XDG_DATA_DIRS` is set)
- `usr{,/local}/share/openxcom` (if `$XDG_DATA_DIRS` isn't set)
- `${CMAKE_INSTALL_FULL_DATADIR}/openxcom`
- `./`

#### What actually happens:
Data files get installed to `${CMAKE_INSTALL_FULL_DATADIR}/openxcom`
The game searches the following locations for data files:
- `$XDG_DATA_HOME/openxcom` (if `$XDG_DATA_HOME` is set)
- `$HOME/.local/share/openxcom` (if `$XDG_DATA_HOME` isn't set)
- `$XDG_DATA_DIRS/openxcom` (if `$XDG_DATA_DIRS` is set)
- `usr{,/local}/share/openxcom` (independently of `$XDG_DATA_DIRS`)
- `${CMAKE_INSTALL_FULL_DATADIR}/openxcom`
- `./`

The two commits responsible are #474 which added `${DATADIR}` and also put `usr{,/local}/share/openxcom` on the directory list independently of the state of `$XDG_DATA_DIRS`, and #1057, which then replaced `${DATADIR}` `${CMAKE_INSTALL_FULL_DATADIR}/openxcom` causing `${DATADIR}` to go unused.

I should also mention that while the desription of `${DATADIR}` is wrong, so was my understanding of it. It is described as "Where to place datafiles", while I thought it meant "Where to look for datafiles".

(`$VAR` means run-time environment variable, `${VAR}` means build-time CMake variable)

mcz:
-DDATADIR used to have no effect. Now it adds a directory to the data search path.
Also changes the system-wide data directories (/usr/share, /usr/local/share, /Users/Shared/OpenXcom) to only be searched if XDG_DATA_DIRS is unset or empty.

fixes #1391


CMakeLists.txt
```diff
@@ -19,7 +19,7 @@ option ( ENABLE_CLANG_ANALYSIS "When building with clang, enable the static anal
 option ( CHECK_CCACHE "Check if ccache is installed and use it" OFF )
 set ( MSVC_WARNING_LEVEL 3 CACHE STRING "Visual Studio warning levels" )
 option ( FORCE_INSTALL_DATA_TO_BIN "Force installation of data to binary directory" OFF )
-set ( DATADIR "" CACHE STRING "Where to place datafiles" )
+set ( DATADIR "" CACHE STRING "Where to search for datafiles" /)
```
SupSuper, line 22: Typo?

src/CMakeLists.txt
```diff
@@ -431,9 +431,12 @@ set ( openxcom_src ${root_src} ${basescape_src} ${battlescape_src} ${engine_src}
 set ( install_dest RUNTIME )
 set ( set_exec_path ON )
 set ( install_dest_dir bin )
+if ( NOT "${DATADIR}" STREQUAL "" )
+  add_definitions( -DDATADIR="${DATADIR}" )
+endif ()
 if ( UNIX AND NOT APPLE )
   set ( data_install_dir "${CMAKE_INSTALL_FULL_DATADIR}/openxcom" )
-  add_definitions( -DDATADIR="${data_install_dir}/" )
+  add_definitions( -DINSTALLDIR="${data_install_dir}/" )
```
SupSuper, line 439: Why the separate define?

src/Engine/CrossPlatform.cpp
```diff
@@ -224,16 +230,17 @@ std::vector<std::string> findDataFolders()
 			dir = strtok(0, ":");
 		}
 	}
+	else
+	{
```
SupSuper, line 234: Won't this break use cases where data is spread among multiple folders?

CMakeLists.txt
mcz: The slash? Yes, sorry. I fixed it.

src/CMakeLists.txt
mcz: So the executable knows where its data files were installed. Currently passing DATADIR to cmake does nothing and instead CMAKE_INSTALL_FULL_DATADIR is used to pass -DDATADIR to the preprocessor. Instead of removing the DATADIR cmake option I repurposed it to add an entry to the directory list. A possible application is wanting to install the game files to /usr/share but searching for data files in /opt/ as well (e.g. from a commercial copy of xcom)

As for changing the name of the define, I thought it would be confusing if DATADIR (the C #define) would be set by CMAKE_INSTALL_FULL_DATADIR (the cmake variable) while DATADIR (the cmake variable) was used to set a different c define. This way DATADIR sets DATADIR.

src/Engine/CrossPlatform.cpp
mcz: It will only break use cases where all of the following apply:
- The data in /usr/(local/)share
-  /usr/(local/)share isn't in $XDG_DATA_DIRS
- The game wasn't installed to /usr/(local/)share
- DATADIR was not set to /usr/(local/)share

/usr/(local/)share are supposed to be only a fallback, if XDG_DATA_DIRS isn't set:
> $XDG_DATA_DIRS defines the preference-ordered set of base directories to search for data files in addition to the $XDG_DATA_HOME base directory. The directories in $XDG_DATA_DIRS should be seperated with a colon ':'.
>
>If $XDG_DATA_DIRS is either not set or empty, a value equal to /usr/local/share/:/usr/share/ should be used. 

(https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)

PR #1401, opened 2023-01-06
Prevent lines going missing when trying to scroll when not needed.

R1dO:
This PR causes `TextList::scrollTo()` to return early when scrolling is not needed.

This should prevent a 'bug' where lines go missing when programmatically trying to scroll lists smaller than allocated for.

-- edit --
Please disregard last line of commit message. I've misinterpreted original behavior.
Unintentional hiding of lines due to scroll still seems like an error to me though.


R1dO:
In case one was wondering about the force-push.

Original commit had an error in the commit message, the force-push removed that line.

PR #1407, opened 2023-02-20
Fix loading civilianTypes

MeridianOXC:
https://openxcom.org/forum/index.php/topic,11109.0.html


MeridianOXC:
PR 1409 already takes care of this.

PR #1427, opened 2023-09-15
fix: remove discrepancy between displayed and real accuracy

narical:
Code for displayed accuracy (for UFO Extender option) differs from the actual accuracy calculation inside Projectile::applyAccuracy(), with different rounding. What's more important, displayed number in Map::drawTerrain() is calculated in 3D space, and applyAccuracy ignores z-coordinates whatsoever, so displayed number is actually more precise. This fix makes those numbers in line with each other.

Changes aren't "pure cosmetic", especially with corner cases - game won't consider target which is 8 levels above targeting unit, as adjanced for accuracy calculation.

### DIFF
--- a/src/Battlescape/Projectile.cpp
+++ b/src/Battlescape/Projectile.cpp
@@ -335,7 +335,8 @@ void Projectile::applyAccuracy(Position origin, Position *target, double accurac
 {
 	int xdiff = origin.x - target->x;
 	int ydiff = origin.y - target->y;
-	double realDistance = sqrt((double)(xdiff*xdiff)+(double)(ydiff*ydiff));
+	int zdiff = origin.z - target->z;
+	double realDistance = std::ceil( sqrt( (double)(xdiff*xdiff + ydiff*ydiff + zdiff*zdiff) ));
 	// maxRange is the maximum range a projectile shall ever travel in voxel space
 	double maxRange = keepRange?realDistance:16*1000; // 1000 tiles
 	maxRange = _action.type == BA_HIT?46:maxRange; // up to 2 tiles diagonally (as in the case of reaper v reaper)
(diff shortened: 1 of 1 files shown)


MeridianOXC:
1/ both calculations are done in 2D, please read the code carefully

2/ rounding is a factor, but it is negligible... please note that the two places are anyway calculating fundamentally different things... in Map.cpp we calculate a distance between the origin tile and a target tile, whereas in Projectile.cpp we calculate a distance between the origin voxel and a target voxel. They will not return same results even if rounding is changed.

3/ lastly, if we wanted to change anything here (and we most likely don't), we would change the displayed number to match the calculated number, not the other way around.

I recommend closing this PR.

narical:
> both calculations are done in 2D, please read the code carefully
You're right, I didn't notice "considerZ" flag and came to a wrong conclusion.

I'm making my own accuracy-related mod, and was trying to debug inconsistency between displayed and internal accuracy. As it turns out, that inconsistency appears before my own code (that's why I made this PR in the first place). Seems like the difference is due to different distance calculation and (as part of it) different rounding.

> please note that the two places are anyway calculating fundamentally different things
As a player, when I see "ufo extender" accuracy number, I suppose that number is real. In the current state, that number is sometimes real, sometimes not. We're talking about 2% here, and for low-accuracy shots that's a meaningful amount.

> They will not return same results even if rounding is changed.
It depends on what you're trying to achieve. If your goal is to get consistent numbers so player can make a well-informed desision, it's totally possible to change calculations to make them in line, either per-voxel or per-tile basis.

There's a huge point in your response tho... It basically means the game totally ignores Z-coordinate, and considers two distances (to the adjanced unit / to the 7 levels higher one) as totally equal, whereas the second distance is tenfold larger in 16x16 tile equivalent.  If that's the case and it's intentional and represent vanilla x-com experince - there's nothing one can do about, it's a bug which couldn't be "fixed" in upstream version.

MeridianOXC:
> It depends on what you're trying to achieve. If your goal is to get consistent numbers so player can make a well-informed desision, it's totally possible to change calculations to make them in line, either per-voxel or per-tile basis.

IMO, only voxel-based (for both) would be counted as an improvement. Tile-based (for both) would be less precise and non-vanilla.

I don't remember all the details of the origin/target voxel calculation, but it's certainly non-trivial and would need a big change in the display calculations.

And to put the final nail in the coffin, some of the calculations even use RNG (target voxel calculations for arcing trajectories I think?), making it (strictly speaking) impossible to have a 100% consistent result in every case.

PR #1429, opened 2023-10-10
Update Polish.nam

DerDoktor82:
Add Grzegorz Brzęczyszczykiewicz

### DIFF
--- a/bin/common/SoldierName/Polish.nam
+++ b/bin/common/SoldierName/Polish.nam
@@ -34,6 +34,7 @@ maleFirst:
   - Jacek
   - Janusz
   - Miroslaw
+  - Grzegorz
 femaleFirst:
   - Anna
   - Maria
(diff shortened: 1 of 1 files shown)


MeridianOXC:
The file already contains the name Grzegorz. And what's the rationale behind adding the name Brzęczyszczykiewicz to xcom?

DerDoktor82:
oh sorry. my bad about Grzegorz.
Grzegorz Brzęczyszczykiewicz is a famous polish "fun" name. It was used in a movie long time ago. It is that special because its hard to pronounce, especially for germans.
Thought it was a funny addition and i wanted to contribute something.

MeridianOXC:
The thing is, it's too long. Only the shortest first name "Jan" fits on the screen together with this surname. All other combinations will be cut off.

DerDoktor82:
ah i see. thank you. in that case lets revoke this case. 
Have a nice day.

PR #1436, opened 2024-03-23
Add auto-chase-usos and initial-intercept-stance options

Marshallx:
Adds 2 quality of life options around intercepting -

"Auto chase USOs" (TFTD)
When intercepting a USO over land, the game invariably says "cannot engage while airborne / continue pursuit?"
This gets really annoying when 9 times out of 10 you still want to pursue the USO until you can engage it.
This option suppresses the dialog and continues to chase the USO until your craft can engage it.

"Initial intercept stance"
By default, all dogfights start in "standoff" mode. You have to choose an attack stance manually. In some cases this wastes precious time before a UFO will flee.
This option allows the user to automatically start dogfights in their preferred stance (personally I like to change this to "cautious")

Also fixed a couple of minor typos (plural of craft is still craft not crafts)


bin/common/Language/en-US.yml
BitterlyIronic, line 1: STR_INITIALINTERCEPTSTANCE and STR_INITIALINTERCEPTSTANCE_DESC are missing from this file

Marshallx: Thanks for spotting that, fixed.

BitterlyIronic (approved):

PR #1444, opened 2024-08-12
Make globe radar circles more visible

Yankes:
Fix in visibility when earth shade is close to shade value of radar circle.


MeridianOXC:
Abandoned.

PR #1486, opened 2026-02-07
Update cpp standard to 20

luqasz:
Fixes `now()` function buffer overflow issue.

### DIFF
--- a/src/Engine/CrossPlatform.cpp
+++ b/src/Engine/CrossPlatform.cpp
@@ -222,8 +222,8 @@ std::vector<std::string> findDataFolders()
 	char const *const xdg_data_dirs = getenv("XDG_DATA_DIRS");
 	if (xdg_data_dirs && *xdg_data_dirs)
 	{
-		char xdg_data_dirs_copy[strlen(xdg_data_dirs)+1];
-		strcpy(xdg_data_dirs_copy, xdg_data_dirs);
+		char xdg_data_dirs_copy[MAXPATHLEN];
+		strncpy(xdg_data_dirs_copy, xdg_data_dirs, MAXPATHLEN);
 		char *dir = strtok(xdg_data_dirs_copy, ":");
 		while (dir != 0)
 		{
@@ -1169,23 +1169,23 @@ void stackTrace(void *ctx)
  */
 std::string now()
 {
-	const int MAX_LEN = 25, MAX_RESULT = 80;
+	const std::size_t DT_MAX_LEN = 25;
+	const std::size_t MAX_RESULT = 80;
 	char result[MAX_RESULT] = { 0 };
 #ifdef _WIN32
-	char date[MAX_LEN], time[MAX_LEN];
-	if (GetDateFormatA(LOCALE_INVARIANT, 0, 0, "dd'-'MM'-'yyyy", date, MAX_LEN) == 0)
+	char date[DT_MAX_LEN], time[DT_MAX_LEN];
+	if (GetDateFormatA(LOCALE_INVARIANT, 0, 0, "dd'-'MM'-'yyyy", date, DT_MAX_LEN) == 0)
 		return "00-00-0000";
-	if (GetTimeFormatA(LOCALE_INVARIANT, TIME_FORCE24HOURFORMAT, 0, "HH'-'mm'-'ss", time, MAX_LEN) == 0)
+	if (GetTimeFormatA(LOCALE_INVARIANT, TIME_FORCE24HOURFORMAT, 0, "HH'-'mm'-'ss", time, DT_MAX_LEN) == 0)
 		return "00-00-00";
-	sprintf(result, "%s_%s", date, time);
+	snprintf(result, MAX_RESULT,"%s_%s", date, time);
 #else
-	char buffer[MAX_LEN];
+	char buffer[DT_MAX_LEN];
 	time_t rawtime;
 	struct tm *timeinfo;
 	time(&rawtime);
 	timeinfo = localtime(&rawtime);
-	strftime(buffer, MAX_LEN, "%d-%m-%Y_%H-%M-%S", timeinfo);
-	sprintf(result, "%s", buffer);
+	strftime(result, MAX_RESULT, "%d-%m-%Y_%H-%M-%S", timeinfo);
 #endif
 	return result;
 }
--- a/src/Engine/CrossPlatform.cpp
+++ b/src/Engine/CrossPlatform.cpp
@@ -222,8 +222,8 @@ std::vector<std::string> findDataFolders()
 	char const *const xdg_data_dirs = getenv("XDG_DATA_DIRS");
 	if (xdg_data_dirs && *xdg_data_dirs)
 	{
-		char xdg_data_dirs_copy[MAXPATHLEN];
-		strncpy(xdg_data_dirs_copy, xdg_data_dirs, MAXPATHLEN);
+		char xdg_data_dirs_copy[strlen(xdg_data_dirs)+1];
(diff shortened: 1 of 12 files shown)


MeridianOXC:
Hi, where exactly is this buffer overflow issue?
I fail to see it.
How can I test it?

src/Engine/CrossPlatform.cpp
```diff
@@ -222,8 +222,8 @@ std::vector<std::string> findDataFolders()
 	char const *const xdg_data_dirs = getenv("XDG_DATA_DIRS");
 	if (xdg_data_dirs && *xdg_data_dirs)
 	{
-		char xdg_data_dirs_copy[strlen(xdg_data_dirs)+1];
-		strcpy(xdg_data_dirs_copy, xdg_data_dirs);
```
luqasz, line 226: @MeridianOXC Buffer overflow. `strcpy` instead of `strncpy`

src/Engine/CrossPlatform.cpp
```diff
@@ -1169,23 +1169,23 @@ void stackTrace(void *ctx)
...
-	char date[MAX_LEN], time[MAX_LEN];
-	if (GetDateFormatA(LOCALE_INVARIANT, 0, 0, "dd'-'MM'-'yyyy", date, MAX_LEN) == 0)
+	char date[DT_MAX_LEN], time[DT_MAX_LEN];
+	if (GetDateFormatA(LOCALE_INVARIANT, 0, 0, "dd'-'MM'-'yyyy", date, DT_MAX_LEN) == 0)
 		return "00-00-0000";
-	if (GetTimeFormatA(LOCALE_INVARIANT, TIME_FORCE24HOURFORMAT, 0, "HH'-'mm'-'ss", time, MAX_LEN) == 0)
+	if (GetTimeFormatA(LOCALE_INVARIANT, TIME_FORCE24HOURFORMAT, 0, "HH'-'mm'-'ss", time, DT_MAX_LEN) == 0)
 		return "00-00-00";
-	sprintf(result, "%s_%s", date, time);
+	snprintf(result, MAX_RESULT,"%s_%s", date, time);
```
luqasz: @MeridianOXC `sprintf` instead of `snprintf`

src/Engine/CrossPlatform.cpp
Yankes: How exactly? Buffer have already all needed space.

src/Engine/CrossPlatform.cpp
Yankes: All buffers are fixed max size, its impossible to have buffer overflow here

luqasz: If you change `MAX_RESULT` in future to smaller then you can have buffer overflow. At least `snprintf` won't allow it.

Yankes: Then there is no buffer overflow and this change is pointless. Nobody will touch this code in next decade probably anyway.

luqasz: I can remove changes in this file from this PR. I can address them separately in future.

luqasz:
I've made changes in `src/CMakeFiles/openxcom.dir/Engine/CrossPlatform.cpp` due to warnings. Since they are warnings and not hard errors, I will address them in separate PR.
```cpp
ninja: Entering directory `build_mac'
[64/172] Building CXX object src/CMakeFiles/openxcom.dir/Engine/CrossPlatform.cpp.o
/Users/lkostka/workspace/oss/OpenXcom/src/Engine/CrossPlatform.cpp:225:27: warning: variable length arrays in C++ are a Clang extension [-Wvla-cxx-extension]
  225 |                 char xdg_data_dirs_copy[strlen(xdg_data_dirs)+1];
      |                                         ^~~~~~~~~~~~~~~~~~~~~~~
/Users/lkostka/workspace/oss/OpenXcom/src/Engine/CrossPlatform.cpp:225:34: note: initializer of 'xdg_data_dirs' is not a constant expression
  225 |                 char xdg_data_dirs_copy[strlen(xdg_data_dirs)+1];
      |                                                ^
/Users/lkostka/workspace/oss/OpenXcom/src/Engine/CrossPlatform.cpp:222:20: note: declared here
  222 |         char const *const xdg_data_dirs = getenv("XDG_DATA_DIRS");
      |                           ^
/Users/lkostka/workspace/oss/OpenXcom/src/Engine/CrossPlatform.cpp:1188:2: warning: 'sprintf' is deprecated: This function is provided for compatibility reasons only.  Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead. [-Wdeprecated-declarations]
 1188 |         sprintf(result, "%s", buffer);
      |         ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/_stdio.h:278:1: note: 'sprintf' has been explicitly marked deprecated here
  278 | __deprecated_msg("This function is provided for compatibility reasons only.  Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead.")
      | ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/sys/cdefs.h:227:48: note: expanded from macro '__deprecated_msg'
  227 |         #define __deprecated_msg(_msg) __attribute__((__deprecated__(_msg)))
```

Yankes:
Ok, removing warnings is more reasonable than "buffer overflow", but `snprintf` is not solution as it cut output that make result bugged, overall better would be change code that we do not need `snprintf` or ``sprintf`` at all.
Right now I work on `[strlen(xdg_data_dirs)+1];` as it not legal C++ code and only extension in GCC and other compilers.

ihhub:
The current code base doesn't even use most of C++11 features. Upgrading to C++20 brings no value at all. Moreover, it would make compilation much-much slower. The project needs to use most of C++11 features and at very best upgrade to C++17 standard to utilize `constexpr` and `auto` stuff. The newest standards of C++ aren't needed for this code.

luqasz:
I haven't noticed any significant increase in compilation time with C++20. C++ is already well established in e.g. linux distributions and won't cause any "missing C++20 standard" compiler. 

Upgrading to C++20 doesn't introduce any compatibility errors with libraries. However sticking to older standards may cause problems in future, when e.g. some library (which is used by this project) deprecates and finally drops older C++ standards.

ihhub:
> I haven't noticed any significant increase in compilation time with C++20. C++ is already well established in e.g. linux distributions and won't cause any "missing C++20 standard" compiler.
> 
> Upgrading to C++20 doesn't introduce any compatibility errors with libraries. However sticking to older standards may cause problems in future, when e.g. some library (which is used by this project) deprecates and finally drops older C++ standards.

C++ compilers **always** backward compatible so using old C++ standard is absolutely fine. The project uses already deprecated SDL 1 library which is not maintained anymore. Plus yaml library which is very outdated. So, there is no reason to even worry about using the older C++ standard for these libraries.

C++20 doesn't bring any features this project can benefit from. Also, Visual Studio 2010 doesn't support C++20. It have limited C++11 support.

Unless, there is a really valid reason to do this, there is no real reason for the upgrade but it's up to the developers of this repo.

Yankes:
C++20 is "slower" as it add more types and other helpers to headers but its only visible when only things that you compile is empty `int main()`. At least in OXCE it could be dwarfed by scripts header.

But allowing better compatibility for C++20 is long term benefit for project. I do not know if basic OXC grain anything from it but OXCE will.

One case is remove of `char xdg_data_dirs_copy[strlen(xdg_data_dirs)+1];` by better C++20 code.

